PR #4, opened 2016-05-04
hstspreload server review by meacer@

lgarron:
meacer@, could you review these changes?

See the [hstspreload v2 doc](https://docs.google.com/document/d/17kft9NgnqdKJTMbt9hbvCSd2-k9T76FP7C_IrM-kisQ/edit#) for an overview of the new code.

(See https://github.com/chromium/hstspreload/pull/66 for the Go library review.)


database.go
```diff
@@ -0,0 +1,181 @@
...
+	Message string `datastore:",noindex" json:"message,omitempty"`
+	// The Unix time this domain was last submitted.
+	SubmissionDate time.Time `json:"-"`
+}
+
+// Updates the given domain updates in batches.
+// Writes and flushes updates to w.
+func putStates(updates []DomainState, statusReport func(format string, args ...interface{})) error {
+	if len(updates) == 0 {
+		statusReport("No updates.\n")
```
meacer: Are you formatting the code with gofmt? (asking because the indendation seems different from google internal styles, but I don't know if it's the default)

files/index.html
```diff
@@ -0,0 +1,124 @@
...
+  <meta name="viewport" content="width=device-width, initial-scale=0.5">
+  <title>HSTS Preload Submission</title>
+  <link rel="stylesheet" href="static/style.css">
+  <script src="static/index.js"></script>
+</head>
+<body class="theme-green">
+
+<div class="content first-content">
+
+
```
meacer: nit: Remove extra blank line

files/index.html
```diff
@@ -0,0 +1,124 @@
...
+  <div class="form">
+
+    <form id="domain-form">
+
+      <h2><label for="domain">
+        Enter a domain for the HSTS preload list:
+      </label></h2>
+
+      <input id="domain" type="text" placeholder="example.com" autofocus autocorrect="off" autocapitalize="off" spellcheck="false"></input>
```
meacer: nit: You might want to wrap these lines at 80 characters. I know go doesn't have a line length limit, but html could have one :)

meacer: Also, why turn off autocomplete? Sounds like some developers might try the same domain multiple times.

files/index.html
```diff
@@ -0,0 +1,124 @@
...
+
+</div>
+
+<div class="content submission">
+
+  <div class="inner-content">
+
+    <h2>
+      Information
+    </h2>
```
meacer: nit: Fits in a single line

files/index.html
```diff
@@ -0,0 +1,124 @@
...
+
+    <div id="text">
+      <p>This form is used to submit domains for inclusion in Chrome's <a href="https://en.wikipedia.org/wiki/HTTP_Strict_Transport_Security">HTTP Strict Transport Security (HSTS)</a> preload list. This is a list of sites that are hardcoded into Chrome as being HTTPS only. <a href="https://blog.mozilla.org/security/2012/11/01/preloading-hsts/">Firefox</a>, Safari, <a href="https://blogs.windows.com/msedgedev/2015/06/09/http-strict-transport-security-comes-to-internet-explorer-11-on-windows-8-1-and-windows-7/">IE 11 and Edge</a> also have HSTS preload lists which include the Chrome list. (See the <a href="http://caniuse.com/#feat=stricttransportsecurity">HSTS compatibility matrix</a>.)</p>
+
+    <h2>
+      Requirements
+    </h2>
```
meacer: Same as above, single line? (In general you'll want to limit the vertical extent of the code)

files/static/index.js
```diff
@@ -0,0 +1,340 @@
...
+      var oReq = new XMLHttpRequest();
+      oReq.addEventListener("load", (function(ev) {
+        resolve(JSON.parse(oReq.response));
+      }).bind(this));
+      oReq.open("GET", "/status/" + domain);
+      oReq.send();
+    });
+  },
+
+  checkdomain: function(domain) {
```
meacer: nit: Camel case (checkDomain)

files/static/index.js
```diff
@@ -0,0 +1,340 @@
...
+  },
+
+  submit: function(domain) {
+    console.log("checkdomain:" + domain);
+    return new Promise(function(resolve, reject) {
+      var oReq = new XMLHttpRequest();
+      oReq.addEventListener("load", (function(ev) {
+        resolve(JSON.parse(oReq.response));
+      }).bind(this));
+      oReq.open("GET", "/submit/" + domain);
```
meacer: GET method shouldn't change state on the server, you should use POST instead.

Also it seems like there is no XSRF protection on these URLs, sound like you'll want it at least for POST urls.

files/static/index.js
```diff
@@ -0,0 +1,340 @@
...
+  },
+
+  clearResults: function() {
+    $("#result").classList.add("hidden");
+  },
+
+  showResults: function(domain, checkDomainIssues, domainStatus) {
+    if (checkDomainIssues.errors.length === 0) {
+      if (checkDomainIssues.warnings.length === 0) {
+        $("#summary").textContent = "" + domain + " satisfies all requirements for preloading!";
```
meacer: Why empty string before domain? Isn't domain always a string? If not, coercing it to a string will mask bugs.

files/static/index.js
```diff
@@ -0,0 +1,340 @@
...
+        $("#summary").textContent = "" + domain + " has " + (checkDomainIssues.warnings.length == 1 ? "a warning" : "warnings") + ", but satisfies all requirements for preloading.";
+        this.setTheme("theme-yellow");
+      }
+    } else {
+        $("#summary").textContent = "" + domain + " has errors and cannot be preloaded.";
+        this.setTheme("theme-red");
+    }
+
+    switch (domainStatus.status) {
+    case "unknown":
```
meacer: I think google style guide wants |case|s to be indented with two spaces. The statements below case don't need to be indented. Example:

switch (...) {
  case 0:
  code0();

  case 1:
  code1();
}

files/static/index.js
```diff
@@ -0,0 +1,340 @@
...
+  setTheme: function(theme) {
+    this.clearTheme();
+    document.body.classList.add(theme)
+  },
+
+  currentDomain: function() {
+    var domain = $("#domain").value;
+    // Check for pasted URLs beginning with http:// or https://
+    if (domain.startsWith("http://")) {
+      domain = domain.slice("http://".length)
```
meacer: early return here and below.

files/static/index.js
```diff
@@ -0,0 +1,340 @@
...
+      }
+    } else {
+        $("#summary").textContent = "" + domain + " has errors and cannot be preloaded.";
+        this.setTheme("theme-red");
+    }
+
+    switch (domainStatus.status) {
+    case "unknown":
+      if (checkDomainIssues.errors.length === 0) {
+        $("#status").textContent = "" + domain + " is not yet preloaded, and may be submitted to the preload list."
```
meacer: Is it possible for this to be true while checkDomainIssues.errors and checkDomainIssues.warnings to be non-empty? If yes, you'll give conflicting messages to the user. Maybe add an assert?

files/static/index.js
```diff
@@ -0,0 +1,340 @@
...
+      } else {
+        $("#status").textContent = "" + domain + " is not preloaded."
```
meacer: I'm finding multiple $("#status").textContent statements a bit confusing. Sounds like you'll want to split this into a new method and return a string here.

database.go
lgarron: Yeah, I have my editor set to run `go fmt` on save. I don't agree with everything, but at least I don't have to worry about it. :-)

files/index.html
lgarron: I also found this confusing at first. However, I found that this set of attributes turns off autocompleting domains *as if they were words*, but still allows a dropdown autocomplete of values you've entered in the field in the past.

meacer: Huh, odd. But if it works, no problem.

files/index.html
```diff
@@ -0,0 +1,124 @@
...
+      <br>
+      <input id="submit" type="submit" disabled
+         value="Submit to the HSTS Preload List">
+    </form>
+
+    <div id="submit-result" class="hidden">
+    </div>
+
+  </div>
+
```
meacer: nit^2: I'd get rid of the extra blank lines here, but really up to you.

files/static/index.js
```diff
@@ -0,0 +1,340 @@
...
+      }
+      break;
+    case "pending":
+      $("#status").textContent = "" + domain + " is pending submission to the preload list."
```
meacer: In here and elsewhere, prefer single quotes over doubles: https://google.github.io/styleguide/javascriptguide.xml?showone=Strings#Strings

files/static/index.js
```diff
@@ -0,0 +1,340 @@
...
+        $("#status").textContent = "" + domain + " was rejected from the preload list for the following reason: " + domainStatus.message;
+      } else {
+        $("#status").textContent = "" + domain + " was rejected from the preload list.";
+      }
+      if (checkDomainIssues.errors.length === 0) {
+        $("#status").textContent += " It may be submitted again."
+      }
+      break;
+    case "removed":
+      if (checkDomainIssues.warnings.length === 0) {
```
meacer: Any reason not to do if (!checkDomainIssues.warnings) instead? (here and elsewhere)

files/static/index.js
```diff
@@ -0,0 +1,340 @@
...
+      } else {
+        $("#status").textContent = "" + domain + " was previously on the preload list, but has been removed.";
+      }
+      break;
+    default:
+      $("#status").textContent = "Cannot determine preload status.";
+    }
+
+    function createIssueElement(issue, bullet, className) {
+      var el = document.createElement("div");
```
meacer: You might consider using closure templates to avoid manually building the dom.

files/static/index.js
```diff
@@ -0,0 +1,340 @@
...
+      bulletSpan.textContent = bullet + " ";
+      el.appendChild(bulletSpan);
+
+      var summary = document.createElement("span")
+      summary.classList.add("summary");
+      summary.textContent = issue.summary;
+      el.appendChild(summary);
+
+      var br = document.createElement("br")
+      el.appendChild(br);
```
meacer: inline br:
el.appendChild(document.createElement("br"))

files/static/index.js
```diff
@@ -0,0 +1,340 @@
...
+
+      var message = document.createElement("span")
+      message.textContent = issue.message;
+      el.appendChild(message);
+
+      return el;
```
meacer: nit: Might want to rename el to container or something like that.

files/static/index.js
```diff
@@ -0,0 +1,340 @@
...
+    }
+
+    $("#errors").textContent = "";
+    $("#warnings").textContent = "";
+    for (var i in checkDomainIssues.errors) {
+      var error = checkDomainIssues.errors[i];
+      var el = createIssueElement(error, "❌", "error");
```
meacer: nit (more of a question): Hmm, I wonder if you should escape these characters and use \unnnn versions. I can't seem to find any references to it in internal code search.

files/static/index.js
```diff
@@ -0,0 +1,340 @@
...
+  hideWaiting: function() {
+    $("#result-waiting").classList.add("hidden");
+  },
+
+  updateURLHash: function(domain) {
+    console.log(domain);
+    if (domain) {
+      history.replaceState({}, document.title, ".?domain=" + domain);
+    } else {
+      history.replaceState({}, document.title, ".");
```
meacer: Single line:
history.replaceState({}, document.title, domain ? '.?domain='  + domain : '.');

files/static/index.js
```diff
@@ -0,0 +1,340 @@
...
+
+  handleResults: function(domain, checkDomainIssues, domainStatus) {
+    console.log("handleResults:", checkDomainIssues, domainStatus);
+
+    if (domain !== this.currentDomain()) {
+      return;
+    }
+    if (this.currentResultsDomain === domain) {
+      return;
+    } else {
```
meacer: No need for else, can simply fall through.

files/static/index.js
```diff
@@ -0,0 +1,340 @@
...
+
+    if (checkDomainIssues.errors.length === 0 && ["unknown", "rejected", "removed"].indexOf(domainStatus.status) != -1) {
+      this.domainToSubmit = domain;
+      this.showSubmission();
+    }
+  },
+
+  domainFromURLParam: function() {
+    var match = window.location.search.match(/^\?domain=([^?&]+)$/);
+    return match ? decodeURIComponent(match[1]) : null;
```
meacer: I haven't followed this thoroughly, but hopefully there is no XSS risk here getting a parameter from the url.

files/static/index.js
```diff
@@ -0,0 +1,340 @@
...
+      $("#domain").value = domain;
+      this.onDomainChanged();
+    }
+  }
+
+  // showResult: function(domain, status) {
+  //   console.log(status);   
+
+  //   this.showElement("resultError");
+  //   $("#resultSummary").textContent = domain + "hi there"
```
meacer: hi

files/static/style.css
```diff
@@ -0,0 +1,234 @@
+html, body {
+  width: 100%;
+}
+
+body {
+  margin: 0;
+  background-color: #f5f5f5;
+  font-family: Helvetica, "Roboto", Tahoma, sans-serif;
```
meacer: Did you mean to have Roboto first? It sounds a bit off to have different fonts in the body vs headers.

files/static/style.css
```diff
@@ -0,0 +1,234 @@
...
+
+  line-height: 24px;
+}
+
+.first-content {
+  margin-top: 4em;
+}
+
+@media (max-width: 839px) {
+  .content{
```
meacer: space before curly brace

server.go
```diff
@@ -0,0 +1,273 @@
...
+
+	http.HandleFunc("/robots.txt", http.NotFound)
+	http.HandleFunc("/favicon.ico", http.NotFound)
+
+	http.HandleFunc("/checkdomain/", checkdomain)
+	http.HandleFunc("/status/", status)
+
+	http.HandleFunc("/submit/", submit)
+	http.HandleFunc("/pending", pending)
+	http.HandleFunc("/update", update)
```
meacer: Should you add a trailing slash to pending and update?

meacer: Also I gather that main() is only used when running locally and appspot handles this automatically. Is that right?

server.go
```diff
@@ -0,0 +1,273 @@
...
+
+	b, err := json.MarshalIndent(v, "", "  ")
+	if err != nil {
+		msg := fmt.Sprintf("Internal error: could not format JSON. (%s)\n", err)
+		http.Error(w, msg, http.StatusInternalServerError)
+		return
+	}
+
+	fmt.Fprintf(w, "%s\n", b)
+	return
```
meacer: no need for return?

server.go
```diff
@@ -0,0 +1,273 @@
...
+}
+
+func status(w http.ResponseWriter, r *http.Request) {
```
meacer: status sounds a bit too generic, maybe getStatus?

server.go
```diff
@@ -0,0 +1,273 @@
...
+		return
+	}
+
+	state.Name = domain
+	writeJSONOrBust(w, state)
+}
+
+func submit(w http.ResponseWriter, r *http.Request) {
+	domainStr := r.URL.Path[len("/submit/"):]
+	domain := chromiumpreload.Domain(domainStr)
```
meacer: |status| function combines these two lines into one. Be consistent.

meacer: Oh I see this is also used below. Never mind then.

server.go
```diff
@@ -0,0 +1,273 @@
...
+			comma = ""
+		}
+
+		fmt.Fprintf(w, `    { "name": "%s", "include_subdomains": true, "mode": "force-https" }%s
+`, name, comma)
+	}
+	fmt.Fprintf(w, "]\n")
+}
+
+func difference(from []chromiumpreload.Domain, take []chromiumpreload.Domain) (diff []chromiumpreload.Domain) {
```
meacer: Some documentation would be nice for this method.

server.go
```diff
@@ -0,0 +1,273 @@
...
+			diff = append(diff, elem)
+		}
+	}
+
+	return diff
+}
+
+func update(w http.ResponseWriter, r *http.Request) {
+	// Get preload list.
+	preloadList, listErr := chromiumpreload.GetLatest()
```
meacer: What happens if two people submit at the same time? It seems like you need some synchronization here (if that's not already the case).

server.go
```diff
@@ -0,0 +1,273 @@
...
+
+	added := difference(actualPreload, databasePreload)
+	for _, name := range added {
+		updates = append(updates, DomainState{
+			Name:   name,
+			Status: StatusPreloaded,
+		})
+	}
+
+	removed := difference(databasePreload, actualPreload)
```
meacer: What happens if the same domain/entry is added and removed at the same time? Is that even possible to force?

files/static/index.js
lgarron: Good point about POST; I've filed a bug: https://github.com/chromium/hstspreload.appspot.com/issues/9

XSRF protection is meaningless in this case because the submission action is purely authenticated on the server; there is no user data or state.

Unless I require submissions to be tied to some sort of account (which I don't plan to do; and even IP addresses are PII), XSRF doesn't prevent anyone from making another person do something they can't do.

lgarron: (The v1 server does have a possible XSRF, but I've taken out the admin functionality that leads to it.)

files/static/index.js
lgarron: Yeah, that was not meant to stay in the code. Removed now.

files/static/index.js
lgarron: Good idea; thanks.

files/static/index.js
lgarron: I just realized clang-format can run on JS. I've formatted the entire project: https://github.com/chromium/hstspreload.appspot.com/commit/47b925c098df563a6aef4512cc8bc1dc2f2ccafc

files/static/index.js
lgarron: Yes. I've overhauled this section to be more clear, although I'm still not perfectly happy with it.

files/static/index.js
lgarron: Yep, done!

files/static/index.js
lgarron: `clang-format` to the rescue!  (But I sure am gonna miss double-quotes. :-( )

files/static/index.js
lgarron: Because it's not correct. ;-)

    ![] === false

files/static/index.js
lgarron: Since there's not so much of it, I'm going taking the DevTools approach of pure JS. I've defined `Element.prototype.createChild` to make things easier to read: https://github.com/chromium/hstspreload.appspot.com/blob/master/files/static/js/view.js#L127

files/static/index.js
lgarron: Hmm, I'd normally be in favor. However, the "." argument is so unusual that I strongly prefer to isolate each case explicitly here.

files/static/index.js
lgarron: Will do.

files/static/index.js
lgarron: Everywhere that handles the domain (both server and JS) is treated as untrusted plain text.

lgarron: That is, in JS I exclusively use `element.textContent` to add text to the DOM (like we do in DevTools).

files/static/style.css
lgarron: Hmm, this originally comes from some Material design templates. I'll move Robot first here.

files/static/style.css
lgarron: Oh man, gotta fix my template. Will do.

server.go
lgarron: Not in this case; `/pending` and `/update` don't take a domain argument.

`main()` is used to start the Go server both locally and on the Managed VM; I believe this is a pretty conventional `main()` function for a server .

server.go
lgarron: Good point. :-D
Will remove.

server.go
lgarron: Go encourages concise names. For the time being, I want to name the handler functions the same as the API paths they handle.

server.go
lgarron: Yeah, this is a bit annoying. :-(

server.go
lgarron: Will do.

server.go
lgarron: (This actually won't run in AppEngine.)

In any case, I believe the state machine for submissions and the idempotency of this function guarantee that there will never be race conditions here. And if there are issues, the next update should take care of it.

(Note that this only updates domains that were previously pending, and have now been added to the Chromium code.)

server.go
lgarron: Nope, not really. ;-)

lgarron:
Thanks for the review. I've updated the code in changes leading up to https://github.com/chromium/hstspreload.appspot.com/commit/5a601e914b001be6f1851f78f060bb01dfce2d25 :-)

lgarron:
Closing for now, but feel free to come back for more. ;-)

server.go
meacer: Hmm okay, the filename (server.go) misled me then. Is it going to run locally?

files/static/index.js
meacer: What's the most appropriate emoji for reactions to JS coercions?

server.go
lgarron: This will be running as a Managed VM, which means the situation (from the perspective of Go code) will be very similar to running locally.

PR #53, opened 2016-07-12
Add observatory.mozilla.org to CORS whitelist

marumari:
The future is soon!  And Thanks!


marumari:
Thanks much for merging this so quickly!

PR #64, opened 2016-10-16
Add look-at-my.site to CORS whitelist

xinan:
Hi, 

I am creating a little app what checks HTTP(S) headers and suggest potential security loopholes for websites. It is at `look-at-my.site`. It is still work-in-progress at the moment, populated with mocks. I would like to add it to the whitelist so that I can use the API for part of my checks. 

Thanks!

### DIFF
--- a/api/cors.go
+++ b/api/cors.go
@@ -20,6 +20,7 @@ var whitelistedHosts = map[string]bool{
 	"apis.infinitudecloud.com": true, // RELEASE
 	"latenight.apis.infinitudecloud.com": true, // BETA
 	"midnight.apis.infinitudecloud.com": true,  // DEVELOPMENT
+	"look-at-my.site": true,
 }
 
 func allowOrigin(clientOrigin string) bool {


xinan:
I signed it!

lgarron:
Thanks for the PR!

Could you give a quick explanation of why your site needs to access the API client-side?

xinan:
One of the checks that our site would like to provide is checking whether a site has HSTS header set, and whether it is in the chrome hardcoded list. If not, we will advice the site owner to do so.

lgarron:
> One of the checks that our site would like to provide is checking whether a site has HSTS header set, and whether it is in the chrome hardcoded list. If not, we will advice the site owner to do so.

Right, in order to check if the HSTS header is set, you'll need to connect from another server (or at least a non-browser client). Is it possible to use the hstspreload API from that server, too?

xinan:
I was actually thinking of using the hstspreload API to check the header. Now we have plans to implement more advanced checks. So I guess we will use a server for that. This PR could be discarded. Thanks!

lgarron:
> Now we have plans to implement more advanced checks. So I guess we will use a server for that. This PR could be discarded. Thanks!

Alright, good luck!

Issue #47, opened 2016-06-21
Allow servers that block port 80 to preload


ghost:
The second requirement says: "Redirect from HTTP to HTTPS on the same host." I'd like to suggest  changing it to "redirect from HTTP to HTTPS on the same host, or don't allow connections to port 80". The reason is that some servers may want to completely block plaintext HTTP if the domain is not visited by users directly (such as image, JavaScript hosting domains).

lgarron:
I think that's reasonable, and I expect we'll implement it in the future.

However, very few sites (like, 1 every few months) ask to be preloaded without allowing connections on port 80. In general, I think it is good to keep enforcing a common best practice.

If you have such a site, I'm happy to verify the other requirements and add it manually for now.

nextPrime:
@lgarron Thank you! No, I don't have such a site to preload. It's just a suggestion that came up to my mind when I saw a website that did so.

lgarron:
Cool.

The tracking bug for this is in the project for the core library; I'll close this bug in favor of that one: https://github.com/chromium/hstspreload/issues/74

lgarron:
Got another email about this today. Let's keep this open as a feature request for automated submission.

lgarron:
Another one today. (Just commenting to keep track of cadence.)

Darkudorus:
And yesterday ;-)

lgarron:
There have been a lot more requests for this as of the last few weeks, so this is now supported.

PR #70, opened 2016-12-12
Clarify that you don't have to listen on HTTP

dcarley:
Per the following:

- chromium/hstspreload.appspot.com#47
- chromium/hstspreload@5c52e6516f3e2171bd6a75d641a2ddf099e36b35

It is no longer necessary to listen on HTTP *just* to serve a redirect to
HTTPS. Re-word the submission requirements to make this clear and match the
underlying library.

### DIFF
--- a/frontend/index.html
+++ b/frontend/index.html
@@ -104,7 +104,7 @@
     <p>In order to be accepted to the HSTS preload list through this form, your site must satisfy the following set of requirements:</p>
     <ol>
       <li>Serve a valid <b>certificate</b>.</li>
-      <li><b>Redirect</b> from HTTP to HTTPS on the same host.</li>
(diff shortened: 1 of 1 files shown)


dcarley:
I signed it!

lgarron:
Thanks. I was going to hold off on this (the phrasing and details of requirements will probably still change over the next few months), but I think it's safe to land this change now.

Would you mind moving the new clause to the end of the sentence, so that the highlighted word remains at the beginning?

dcarley:
> Would you mind moving the new clause to the end of the sentence, so that the highlighted word remains at the beginning?

No problem, I've re-worded it now.

I was also considering using an unordered sub-list, but I'm easy. Feel free to restructure it again in future.

PR #83, opened 2017-02-10
Update caniuse.com link

jsha:
Their internal link structure changed. Also use an HTTPS URL.

### DIFF
--- a/frontend/index.html
+++ b/frontend/index.html
@@ -94,7 +94,7 @@
       This is a list of sites that are hardcoded into Chrome as being HTTPS only.
     </p>
     <p>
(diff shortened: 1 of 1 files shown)


lgarron:
> Also use an HTTPS URL.

caniuse.com doesn't have a valid cert yet. :-/
(But it's the most canonical place for browser compatibility information to link to.)

jsha:
That is weird, I could have sworn I copy-pasted that URL and it worked yesterday. Must have been hallucinating.

lgarron:
Looks like I never got back to this PR after you fixed the scheme. Merged now. :-)

PR #86, opened 2017-04-19
Update index.html: caniuse supports HTTPS

xiaoyinl:
caniuse.com now supports HTTPS, so let's use https://caniuse.com/#feat=stricttransportsecurity.


lgarron:
Thanks!

PR #91, opened 2017-05-28
Convert domain to lowercase during ASCII normalization in the API handlers.

martijnc:
The API currently considers domain.tld and dOmain.tld to be 2 different domains and allows all variations to be preloaded separately.

The domain names should be processed in a case-insensitive way.

The frontend converts domain names to lowercase [1] but the API does not.

[1] https://github.com/chromium/hstspreload.org/blob/master/frontend/static/js/base.js#L31

### DIFF
--- a/api/domain_handlers.go
+++ b/api/domain_handlers.go
@@ -12,6 +12,15 @@ import (
 	"github.com/chromium/hstspreload.org/database"
 )
 
+func normalizeDomain(unicode string) (string, error) {
+	ascii, err := idna.ToASCII(unicode)
+	if err != nil {
+		return "", err
+	}
+
+	return strings.ToLower(ascii), nil
+}
+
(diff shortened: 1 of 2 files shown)


lgarron:
Thanks!

This sounds like a good change.
Would you be able to do two things?

1. Update the commit to be more specific, e.g. "Convert domain to lowercase during ASCII normalization".
2. Add a `domain_handlers_test.go` file with a [table-driven test](https://github.com/golang/go/wiki/TableDrivenTests) similar to the ones [in the core library](https://github.com/chromium/hstspreload/blob/c1c0b7452e2e88a8ed3aa88b3505a50fb3e72b2c/issues_test.go#L24) . Two cases (one starting lowercase, one starting mixed case) should be fine for a start.

PR #93, opened 2017-06-28
Specify lang="en"

coliff:
Add  lang="en" to the HTML page


lgarron:
Thanks for the pull request!

https://www.w3.org/International/questions/qa-lang-why doesn't really give any compelling reasons here, but I'm happy to add this.

PR #97, opened 2017-09-11
Updated APIs

ByJamesBurton:
Removed all 0.me.uk domains and replace with one ncsccs.com entry.

### DIFF
--- a/api/cors.go
+++ b/api/cors.go
@@ -16,8 +16,7 @@ const (
 var whitelistedHosts = map[string]bool{
 	"mozilla.github.io":       true,
 	"observatory.mozilla.org": true,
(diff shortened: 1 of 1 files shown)


lgarron:
I can't access a.ncsccs.com – will it be publicly available?
Could you also tell us more about what you plan to use it for at a.ncsccs.com?

lgarron:
Sounds good, as long as you make clear that your service is not maintained or endorsed by Chrome/Google. :-)

ByJamesBurton:
> Sounds good, as long as you make clear that your service is not maintained or endorsed by Chrome/Google. :-)

Yeah. That's fine.

PR #100, opened 2017-09-20
Increase number of probes.

### LINKED ISSUE #99: Increase waiting for GCD local store
I've gotten the local development setup on my computer (yay!) but I apparently have a slow-ish computer.

The `go run` command polls an HTTP endpoint to see that the local datastore is running:

https://github.com/chromium/hstspreload.org/blob/52a8473b03b2dffb7199bf6b7f6fd664788f4a34/database/gcd/gcd.go#L114-L117

This oddly just isn't enough time for my computer. I would suggest either increasing the number of probes or the time between probes. I set the probe interval to 4x the current value and now I am able to consistently get it up an running.

Thoughts?

vcsjones:
Double the number of probes to prevent premature aborts on waiting for the local GCD environment to spin up.

Fixes #99.

### DIFF
--- a/database/gcd/gcd.go
+++ b/database/gcd/gcd.go
@@ -18,7 +18,7 @@ import (
 )
 
 const (
-	numLocalProbes    = 10
+	numLocalProbes    = 20
 	initialProbeSleep = 500 * time.Millisecond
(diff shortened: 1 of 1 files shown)


lgarron:
Thanks!

PR #101, opened 2017-09-21
Show issues for preloaded domains

### LINKED ISSUE #55: Report requirements status even if a site is preloaded
I had an error in my HSTS header but hstspreload.appspot.com didn't alert me to this error because my site was already in the preload list.

vcsjones:
* The site will now show all warnings and errors for already preloaded domains unless they are bulk domains.
* Introduces a new type called `DomainStateWithBulk`. I would like to come up with a better name than this. I'm also not sure where to put it.
* To keep the changeset small, I introduced a new field on `wantsBody` to handle the bulk status test assertions instead of moving everything to asserting the bulk status.

Fixes #55

### DIFF
--- a/frontend/static/js/form.js
+++ b/frontend/static/js/form.js
@@ -104,17 +104,16 @@ function statusString(status, issues, domain) {
       return 'Status: ' + domain + ' is pending submission to the preload list.';
     case 'preloaded':
       if (status.bulk) {
-        return 'Status: ' + domain + ' is currently preloaded.';
-      } else {
         switch (worstIssues(issues)) {
-          case 'none':
-            return 'Status: ' + domain + ' is currently preloaded.';
           case 'warnings':
+            return 'Status: ' + domain + ' is currently preloaded, but has the following issues:';
           case 'errors':
-            return 'Status: ' + domain + ' is currently preloaded, but missing current requirements.';
+            return 'Status: ' + domain + ' is currently preloaded, but no longer meets the requirements. It may be at risk of removal.';
           default:
-            return 'Status: ' + domain + ' is currently preloaded, but unable to determine if it meets current requirements.';
+            return 'Status: ' + domain + ' is currently preloaded.';
         }
+      } else {
+        return 'Status: ' + domain + ' is currently preloaded.';
       }
       break;
 
@@ -248,7 +247,9 @@ PreloadController.prototype = {
         break;
       case 'preloaded':
         view.setTheme('theme-green');
-        view.showIssues(issues);
+        if (status.bulk) {
+          view.showIssues(issues);
+        }
         break;
       default:
         throw "Unknown status";
--- a/frontend/static/js/form.js
+++ b/frontend/static/js/form.js
@@ -96,14 +96,23 @@ Form.prototype = {
   }
 };
 
-function statusString(status, domain) {
+function statusString(status, issues, domain) {
   switch (status.status) {
     case 'unknown':
       return 'Status: ' + domain + ' is not preloaded.';
     case 'pending':
       return 'Status: ' + domain + ' is pending submission to the preload list.';
     case 'preloaded':
-      return 'Status: ' + domain + ' is currently preloaded.'
+      switch (worstIssues(issues)) {
+        case 'none':
+          return 'Status: ' + domain + ' is currently preloaded.';
+        case 'warnings':
+        case 'errors':
+          return 'Status: ' + domain + ' is currently preloaded, but missing current requirements.';
+        default:
+          return 'Status: ' + domain + ' is currently preloaded, but unable to determine if it meets current requirements.';
+      }
+      break;      
 
     case 'rejected':
       if (status.message) {
@@ -216,7 +225,7 @@ PreloadController.prototype = {
 
   showResults: function(view, domain, issues, status) {
 
-    view.setStatus(statusString(status, domain));
+    view.setStatus(statusString(status, issues, domain));
 
     var showForm = false;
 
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -1,2 +1,3 @@
 /frontend/version
 /api/bulk_preloaded.go
+.DS_Store
```
lgarron: Nit: Put in a separate PR and keep the newline.

frontend/static/js/form.js
```diff
@@ -96,14 +96,27 @@ Form.prototype = {
...
-      return 'Status: ' + domain + ' is currently preloaded.'
+      if (status.bulk) {
+        return 'Status: ' + domain + ' is currently preloaded.';
+      } else {
+        switch (worstIssues(issues)) {
+          case 'none':
+            return 'Status: ' + domain + ' is currently preloaded.';
+          case 'warnings':
+          case 'errors':
+            return 'Status: ' + domain + ' is currently preloaded, but missing current requirements.';
```
lgarron: I would say:

    return 'Status: ' + domain + ' is currently preloaded, but no longer meets the requirements. It may be at risk of removal';

frontend/static/js/form.js
```diff
@@ -96,14 +96,27 @@ Form.prototype = {
...
+          default:
+            return 'Status: ' + domain + ' is currently preloaded, but unable to determine if it meets current requirements.';
```
lgarron: I would actually fold this with `none`, and not mention anything about the failure.

frontend/static/js/form.js
```diff
@@ -96,14 +96,27 @@ Form.prototype = {
...
       return 'Status: ' + domain + ' is pending submission to the preload list.';
     case 'preloaded':
-      return 'Status: ' + domain + ' is currently preloaded.'
+      if (status.bulk) {
+        return 'Status: ' + domain + ' is currently preloaded.';
+      } else {
+        switch (worstIssues(issues)) {
+          case 'none':
+            return 'Status: ' + domain + ' is currently preloaded.';
+          case 'warnings':
```
lgarron: I'd use a separate status:

    return 'Status: ' + domain + ' is currently preloaded, but has the following issues:';

.gitignore
vcsjones: Will do. Removed commit.

frontend/static/js/form.js
```diff
@@ -96,14 +96,27 @@ Form.prototype = {
...
-function statusString(status, domain) {
+function statusString(status, issues, domain) {
   switch (status.status) {
     case 'unknown':
       return 'Status: ' + domain + ' is not preloaded.';
     case 'pending':
       return 'Status: ' + domain + ' is pending submission to the preload list.';
     case 'preloaded':
-      return 'Status: ' + domain + ' is currently preloaded.'
+      if (status.bulk) {
```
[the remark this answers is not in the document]
vcsjones: Ah! It seems I misunderstood the requirement entirely! Doh.

frontend/static/js/form.js
```diff
@@ -235,6 +247,7 @@ PreloadController.prototype = {
         break;
       case 'preloaded':
         view.setTheme('theme-green');
+        view.showIssues(issues);
```
lgarron: Still needs `if (status.bulk)` here, but the rest looks good!

vcsjones: @lgarron Thanks for your patience and supervision!

vcsjones:
@lgarron I'm a little confused about what `bulk` is.

Take for example, `vcsjones.com`. It is preloaded and (I think) it meets all of the preloading requirements, but it is not a bulk loaded domain. If I were to accidentally remove `preload` directive in the header, I wouldn't get a warning about it even though it is preloaded when I came back to the site. Should I?

Am I misunderstanding something?

lgarron:
> Take for example, vcsjones.com. It is preloaded and (I think) it meets all of the preloading requirements, but it is not a bulk loaded domain.

At the moment, bulk entries only include domains submitted to hstspreloard.org (or hstspreload.appspot.com) on or after February 29, 2017:
- https://hstspreload.org/#continued-requirements
- https://cs.chromium.org/chromium/src/net/http/transport_security_state_static.json (line 8749: `// START OF BULK ENTRIES`)

These are all the ones that we can *currently* programmatically identify as being subject to the removal form policy. This may expand once we have properly classified old bulk entries:
https://chromium-review.googlesource.com/c/chromium/src/+/671235

vcsjones:
>These are all the ones that we can currently programmatically identify as being subject to the continued requirements.

Ah OK! Thanks for clearing that up. My understanding of what bulk is:

>At the moment, bulk entries only include domains submitted to hstspreloard.org (or hstspreload.appspot.com) on or after February 29, 2016:

was backward.

Let me know if anything else in the PR needs to be changed.

lgarron:
> Ah OK! Thanks for clearing that up. My understanding of what bulk is:
> 
> At the moment, bulk entries only include domains submitted to hstspreloard.org (or hstspreload.appspot.com) on or after February 29, 2016:
> 
> was backward.

No worries. It's a very underdocumented concept, and we're still trying to figure out exactly what it means.

This PR is all good. As a follow-up, it might be nice to avoid scaring site owners by only pointing out certain errors as likely to lead to removal:
- `header.preloadable.preload.missing` (means the HSTS header is sent, but without the `preload` directive), which makes it eligible for removal right now: https://hstspreload.org/removal/
- `response.no_header`, since those are the most likely candidates for pruning.

Other errors are unlikely to lead to removal, even though the stated policy says they may.
(This isn't because we don't want to enforce the policy for those errors, but because we *can't trust the unauthenticated response*.)

However, we should still point out the other errors (e.g. missing a redirect from HTTP to HTTPS), and I don't have a good way in mind to adapt the UI to this. So, the current approach is fine unless we get a lot of concerned feedback.

lgarron:
I can't deploy right now, but this will make it into the next deployment.

PR #112, opened 2017-10-21
Update index.html

XhmikosR:
* fix validation errors
* remove unneeded end tags
* use double quotes for attributes consistently
* use `strong` instead of `b` which is semantically more correct
* use lowercase `doctype`; same functionality, slightly better gzip compression
* move github.css along with the rest in `head`

### DIFF
--- a/frontend/index.html
+++ b/frontend/index.html
@@ -1,14 +1,16 @@
-<!DOCTYPE html>
+<!doctype html>
 <html lang="en">
 <head>
+  <meta charset="utf-8">
   <meta name="viewport" content="width=device-width, initial-scale=0.5">
   <title>HSTS Preload List Submission</title>
   <link rel="shortcut icon" href="favicon.ico">
-  <link rel="apple-touch-icon" href="static/app-icon.png"/>
-  <link type="application/opensearchdescription+xml" rel="search" href="search.xml"/>
+  <link rel="apple-touch-icon" href="static/app-icon.png">
+  <link rel="search" href="search.xml" type="application/opensearchdescription+xml">
 
   <link rel="stylesheet" href="static/css/style.css">
   <link rel="stylesheet" href="static/css/form.css">
+  <link rel="stylesheet" href="static/css/github.css">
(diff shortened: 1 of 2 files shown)


frontend/index.html
```diff
@@ -37,7 +38,7 @@
   <!-- We un-hide the form using inline JS so that (when JS is enabled)
        it shows in the normal rendering order as if it was never hidden. -->
   <script>document.getElementById("domain-form").classList.remove("hidden");</script>
-  <noscript>Submitting entries to the HSTS preload list via this site requires Javascript.</noscript>
+  <noscript>Submitting entries to the HSTS preload list via this site requires JavaScript.</noscript>
```
lgarron: I personally refuse to capitalize it as `JavaScript`, but I guess the time of reckoning has come for this project. :-(

lgarron:
Do you know the exact difference between `<tt>` and `<code>` formatting?

frontend/index.html
XhmikosR: I didn't do it either in the past, but https://en.wikipedia.org/wiki/JavaScript. So I guess it's the "official" name. :)

XhmikosR:
`tt` is obsolete, see https://developer.mozilla.org/en-US/docs/Web/HTML/Element/tt

lgarron:
> tt is obsolete, see https://developer.mozilla.org/en-US/docs/Web/HTML/Element/tt

Good to know. I'm just wondering, are there difference in formatting or behaviour that I should expect?

XhmikosR:
Nah, as you can see from the MDN link, browsers just display it with a monospace font.

>This element was intended to style text as it would display on a fixed width display, such as a teletype. It probably is more common to display fixed width type using the `<code>` element.

lgarron:
> Nah, as you can see from the MDN link, browsers just display it with a monospace font.

I'm not complete convinced. :-P
But it looks alright on my computer, so let's go ahead. Thanks for the cleanup!

XhmikosR:
I tested it with the evergreen browsers and it looked the same :)

PR #117, opened 2017-10-23
README.md: fix code block and align logo to the center.

XhmikosR:
I assume this was the intended behavior since `<center>` was being used, but that is obsolete so it didn't have the desired effect :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,11 @@
-<center>
-<img src="frontend/static/app-icon.png" width=144>
-</center>
+<p align="center">
+  <a href="https://hstspreload.org/">
+    <img src="frontend/static/app-icon.png" alt="" width="144" height="144">
+  </a>
 
-# `hstspreload.org`
+  <h1 align="center">hstspreload.org</h1>
+</p>
+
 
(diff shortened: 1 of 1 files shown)


XhmikosR:
I assume this was the intended behavior since `<center>` was being used, but that is obsolete so it didn't have the desired effect :)

README.md
```diff
@@ -1,8 +1,11 @@
-<center>
-<img src="frontend/static/app-icon.png" width=144>
-</center>
+<p align="center">
+  <a href="https://hstspreload.org/">
+    <img src="frontend/static/app-icon.png" alt="" width="144" height="144">
```
XhmikosR: Personally I'll use the svg instead of the png.

lgarron:
> I assume this was the intended behavior since <center> was being used, but that is obsolete so it didn't have the desired effect :)

Yeah, I think it worked on a legacy Markdown viewer I was using at the time.

README.md
lgarron: Indeed, that would be a good idea, since we're already using HTML for markup at this point. I'll merge this for now, though.

PR #118, opened 2017-10-25
Minor CSS cleanup and tweaks.

XhmikosR:
* use the native font stack
* use hex color values when possible for consistency and because it's shorter
* use numbered for-weight
* merge a couple of selectors
* remove zero from units
* use `background-color` instead of the `background` shorthand since only the background color is supposed to change
* style consistency changes
* remove unused selectors

### DIFF
--- a/frontend/static/css/form.css
+++ b/frontend/static/css/form.css
@@ -24,7 +24,7 @@ noscript {
   display: block;
   margin: 2em 4em;
   font-size: 1.5em;
-  color: gray;
+  color: #808080;
 }
 
 /******** result-waiting ********/
@@ -48,16 +48,20 @@ noscript {
   margin-bottom: 1em;
 }
 
-@keyframes spinnerRotate
-{
-  from { transform:rotate(0deg); }
-  to { transform:rotate(360deg); }
+@keyframes spinnerRotate {
+  from {
+    transform: rotate(0);
+  }
+
+  to {
+    transform: rotate(360deg);
+  }
 }
 
 /******** result ********/
 
 #result {
-  padding: 1em 0em;
+  padding: 1em 0;
 }
 
 #result .issues {
@@ -69,7 +73,7 @@ noscript {
 }
 
 #result .issues-list div {
-  border-top: 1px solid rgb(127, 127, 127);
+  border-top: 1px solid #7f7f7f;
   /*background: rgba(255, 0, 0, 0.25);*/
   display: block;
   padding: 1em 2em;
@@ -77,7 +81,7 @@ noscript {
 }
 
 #result .issues .summary {
-  font-weight: bold;
+  font-weight: 700;
 }
 
 #result .issues .message {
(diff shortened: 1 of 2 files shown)


frontend/static/css/style.css
```diff
@@ -5,53 +5,33 @@ html, body {
...
-  color: rgb(66, 66, 66);
-  letter-spacing: 0.02em;
-  display:flex;
-  align-items: center;
-  height: 64px;
-  padding: 0 16px 0 40px;
-
-  font-family: "Roboto", Helvetica, Tahoma, sans-serif;
+  background-color: #7f7f7f;
+  font: 14px/1.5 -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, "Helvetica Neue", Helvetica, Arial, sans-serif;
```
lgarron: Could you prioritize Roboto?

frontend/static/css/form.css
```diff
@@ -24,7 +24,7 @@ noscript {
   display: block;
   margin: 2em 4em;
   font-size: 1.5em;
-  color: gray;
```
lgarron: https://developer.mozilla.org/en-US/docs/Web/CSS/color_value shows that `gray` is `#808080`. If you also want to propose non-refactoring changes, would you mind putting them in a separate PR for easier review? Thanks! :-)

frontend/static/css/style.css
XhmikosR: This is the de facto font system stack. We use it also in Bootstrap.

See https://css-tricks.com/snippets/css/system-font-stack/

frontend/static/css/form.css
XhmikosR: This was a mistake, fixed.

frontend/static/css/style.css
lgarron: Alright, that actually looks pretty reasonable. Thanks for the suggestion!

frontend/static/css/style.css
```diff
@@ -93,16 +65,13 @@ a.hash-link:hover {
...
-}
-
 .hidden {
   display: none;
 }
 
 .header-example {
-  background: rgba(0, 0, 0, 0.1);
+  background-color: rgba(0, 0, 0, 0.1);
+  color: #757575;
```
lgarron: This reduces the contrast, right? Could we keep the current color?

XhmikosR: The previous color doesn't have a contrast ration > 4.5 to be WCAG2AA valid.

The difference is subtle, IMO it should be kept like I have it.

XhmikosR: ...but anyway, it's ultimately your call so I'll revert it.

lgarron: The current color is `#424242`. That's darker ⇒ more contrast, right?

XhmikosR: Yup. Must be a bug in the tool I used (HTM_Codesniffer) to check for this.

lgarron:
Thanks! :-)

PR #119, opened 2017-10-25
Minor JavaScript lint fixes.

XhmikosR:
@lgarron: this is sort of untested except for just running it locally without any server.

So, please try it just in case :)

### DIFF
--- a/frontend/static/js/form.js
+++ b/frontend/static/js/form.js
@@ -1,23 +1,25 @@
 'use strict';
 
+/* global $:false, HSTSPreload:false, URLParam:false, PreloadView:false, extractDomain:false */
+
 var Form = function(controller) {
   this._hstsPreload = new HSTSPreload();
   this._controller = new controller(this._hstsPreload);
 
   this._urlParam = new URLParam();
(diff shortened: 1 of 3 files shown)


XhmikosR:
I tested this locally and seems to be fine.

BTW, one thing I noticed is that the list that gets downloaded locally isn't the latest from the live site.

lgarron:
@vcsjones FYI, this will change the syntax but not semantics of the existing JS. TypeScript in the future would still be great. :-D

lgarron:
Thanks! :-)

frontend/static/js/form.js
```diff
@@ -174,65 +176,68 @@ PreloadController.prototype = {
...
   showPreloadEligibility: function(view, domain, issues) {
     switch (worstIssues(issues)) {
       case 'none':
         view.setSummary(
-            'Eligibility: ' + domain +
+          'Eligibility: ' + domain +
             ' is eligible for the HSTS preload list.');
         view.setTheme('theme-green');
         return true;
         break;
```
XhmikosR: These are unreachable BTW.

PR #125, opened 2017-10-25
README.md: Revert to the PNG logo image.

XhmikosR:


lgarron:
Aaargh, this might have been why I used a the PNG before. I don't want viewing the README to trigger a request to GitHub for no good reason. :-/

Could we just revert to the PNG?

XhmikosR:
But the PNG will also trigger a request... where do you think it loads it from? :P

XhmikosR:
And to be precise both of them are 2 requests when one views the README.md file:

1. a 302 redirect
2. the image itself

lgarron:
> But the PNG will also trigger a request... where do you think it loads it from? :P

If you have a Markdown viewer and are viewing from your checkout: your filesystem. ;-)

XhmikosR:
Ah, that is a different case indeed.

As you wish, I'll just revert to the PNG.

XhmikosR:
Done, just squash all of them :)

lgarron:
I suspect this won't work on GitHub either, and we'll just have to revert the previous commit. :-(
Let me try.

lgarron:
Looks like the PNG works!

Perhaps https://github.com/isaacs/github/issues/316 was only partially fixed.

PR #129, opened 2017-11-03
Move static-data/.gitignore in the root .gitignore.

XhmikosR:
Easier and better for maintenance to have everything in one place.

Also, clean up the entries.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,2 @@
 /frontend/version
-/api/bulk_preloaded.go
+/static-data/bulk-preloaded.json
--- a/static-data/.gitignore
+++ b/static-data/.gitignore
@@ -1,2 +1,0 @@
-/bulk-preloaded.json
-/dos.yaml


.gitignore
```diff
@@ -1,2 +1,4 @@
-/frontend/version
 /api/bulk_preloaded.go
```
lgarron: While we're at it, could we clean up and remove this entry?

.gitignore
```diff
@@ -1,2 +1,4 @@
...
+/frontend/version
+/static-data/bulk-preloaded.json
+/static-data/dos.yaml
```
lgarron: This one, too!

lgarron:
Thanks!

We can remove two of these, could you revise this to take them out?

XhmikosR:
@lgarron: done!

lgarron:
Thanks!

PR #130, opened 2017-11-06
Add policy to entries added by the current auto-roller script.

martijnc:
This updates the current auto-roller script so that new entries are assigned the "bulk-1-year" policy.

I'd like to make the policy field required once all entries are assigned a policy, updating the auto-roller now might make the process a little easier.


lgarron:
Thanks!

lgarron:
I'd also kind of like to switch `include_subdomains` and `mode`, but that's less important. :-)

PR #133, opened 2017-11-14
Remove Lucas Garron name

ByJamesBurton:
Referring to the "preload list maintainers" instead of your name. #110


lgarron:
Thanks!

PR #136, opened 2017-12-14
Update update_bulk_preloaded.py to account for subdomain entries

estark37:
https://chromium-review.googlesource.com/#/c/773787/ added a new bulk entry
section ("1-YEAR BULK SUBDOMAIN HSTS ENTRIES") that update_bulk_preloaded.py
doesn't expect. This patch updates update_bulk_preloaded.py to parse that
section and add its contents to the bulk entries list.

### DIFF
--- a/scripts/update_bulk_preloaded.py
+++ b/scripts/update_bulk_preloaded.py
@@ -14,7 +14,9 @@ class State:
   During18WeekBulkEntries, \
   After18WeekBulkEntries, \
   During1YearBulkEntries, \
-  After1YearBulkEntries = range(7)
+  After1YearBulkEntries, \
+  During1YearBulkSubdomainEntries, \
+  After1YearBulkSubdomainEntries = range(9)
 
 def getRawText():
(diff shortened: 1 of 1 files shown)


estark37:
@ericlaw1979 can you review please?

lgarron:
This LGTM as a short-term fix.

A more semantic approach would be to replace `extractBulkEntries()` with:

    def removeCommentLines(rawText):
      processed = ""
      for line in rawText.splitlines():
        if not re.match(r'^ *//.*', line):
          processed += line
      return processed

    def extractBulkEntries(rawText):
      log("Extracting bulk entries...\n")

      parsed_json = json.loads(removeCommentLines(rawText))
      bulk_entries = [entry for entry in parsed_json["entries"] if entry["policy"].startswith("bulk")]
      return bulk_entries

Note that this will `bulk-legacy` entries to be removable through https://hstspreload.org/removal/ , although that was already the plan.

jianglai:
There is an extra print statement in the script. Do we really want that? @ericlaw1979

PR #137, opened 2017-12-17
update_bulk_preloaded.py: Add new 1-year bulk subdomain states

jianglai:
The new entries in the preload list broke this script because the
first line of these entries contains "BULK" in it. Added two new
states to account for it.


jianglai:
Ah, didn't realize there's already a RP for this. Closing it.

Issue #106, opened 2017-10-16
Distinguish between bulk policies on front page


lgarron:
Follow-up to recent changes, as noticed in https://github.com/EFForg/https-everywhere/issues/13147

vcsjones:
This seems like a good idea. The idea being that we could flag certain errors as likely to lead to removal and adjust the UI accordingly, correct?

While I tinker with TypeScript, anything I can do to help here?

lgarron:
> This seems like a good idea. The idea being that we could flag certain errors as likely to lead to removal and adjust the UI accordingly, correct?

Yeah, right now we apply the new 1-year calculation to all old entries (which only have an 18-week requirement): https://github.com/EFForg/https-everywhere/issues/13147#issuecomment-336672380

A short-term solution would be:
- Modify `chromium/hstspreload` to support modifying calculations (or filtering errors) based on a policy enum.
- Apply the `bulk-18-weeks` policy to all domains currently marked as bulk on the server.

In the longer term, we'll need to support at least three different policies (old bulk, 18-week bulk, and 1-year bulk), which is best done by annotating the preload list source. I plan to work on that soon.

PR #142, opened 2018-02-21
Fix wrong error message on Removal page

xiaoyinl:
The "Status" message on https://hstspreload.org/removal/ page is confusing, e.g. "Status: wikipedia.org is currently preloaded, but no longer meets the requirements. It may be at risk of removal."

This message is clearly for the preload page rather than the removal page.

### DIFF
--- a/frontend/static/js/form.js
+++ b/frontend/static/js/form.js
@@ -98,14 +98,14 @@ Form.prototype = {
   }
 };
 
-function statusString(status, issues, domain) {
+function statusString(status, issues, domain, isRemoval) {
   switch (status.status) {
     case 'unknown':
       return 'Status: ' + domain + ' is not preloaded.';
     case 'pending':
       return 'Status: ' + domain + ' is pending submission to the preload list.';
     case 'preloaded':
-      if (status.bulk) {
+      if (status.bulk && !isRemoval) {
         switch (worstIssues(issues)) {
           case 'warnings':
             return 'Status: ' + domain + ' is currently preloaded, but has the following issues:';
@@ -233,7 +233,7 @@ PreloadController.prototype = {
   },
 
   showResults: function(view, domain, issues, status) {
-    view.setStatus(statusString(status, issues, domain));
+    view.setStatus(statusString(status, issues, domain, false));
 
     var showForm = false;
 
@@ -323,7 +323,7 @@ RemovalController.prototype = {
   },
 
   showResults: function(view, domain, issues, status) {
-    view.setStatus(statusString(status, issues, domain));
+    view.setStatus(statusString(status, issues, domain, true));
 
     var showForm = false;
 


xiaoyinl:
@ericlaw1979 Could you please review?

ericlaw1979:
It seems like the error message should only potentially omit "It may be at risk of removal." in this scenario, but showing the other status seems fine?

xiaoyinl:
@ericlaw1979 I think just showing "Status: xxx is currently preloaded" is fine. The code mistakenly treats the error of _removal_ as error of _preloading_. "But no longer meets the requirements. It may be at risk of removal" makes it look like there's an error that may cause the domain to be removed from the preload list, but it's actually fine: the error ("Contains preload directive") prevents the removal, so it's not at risk of removal.

lgarron:
> It may be at risk of removal" make it look like there's an error that may cause the domain to be removed from the preload list, but it's actually fine: the error ("Contains preload directive") prevents the removal, so it's not at risk of removal.

The domain *is* at risk of removal (ignoring #106 in the case of wikipedia.org) – the front page of hstspreload.org [clearly states this](https://hstspreload.org/#continued-requirements):

> You must make sure your site continues to satisfy the submission requirements at all times. Note that removing the preload directive from your header will make your site immediately eligible for the removal form, and that sites may be removed automatically in the future for failing to keep up the requirements.

However, we could certainly do a better job of making clear that it can't be removed in its current state through the web form.

lgarron:
For what it's worth, I would prefer a fix that clarifies that the domain may automatically be removed in the future, but the current state of the PR looks good to me, too.

xiaoyinl:
> The domain is at risk of removal (ignoring #106 in the case of wikipedia.org) – the front page of hstspreload.org clearly states this:

Actually it's not at risk of removal. Wikipedia.org sends `Strict-Transport-Security: max-age=106384710; includeSubDomains; preload`. This is why I think it's confusing.

cristianlivella:
Wow, it's been 8 years, and this still hasn't been merged?
I just ran into the same problem, and the error message is very misleading.

@lgarron the domain is not not at risk of removal. In fact, it's exactly the opposite.

The problem is that [form.js](https://github.com/chromium/hstspreload.org/blob/main/frontend/static/js/form.js) is using the same logic for both the submission form and the removal form.
If the backend returns an "errors" message, this phrase is displayed:
> [domain] is currently preloaded, but no longer meets the requirements. It may be at risk of removal.

This is correct if we are using the submission form, but if we are using the removal form, it is completely wrong, because in that case the endpoint returns an "errors" message when it is not possible to request domain removal, since HSTS is properly enabled in the header.

edit: added API response
```
{
  "errors": [
    {
      "code": "header.removable.contains.preload",
      "summary": "Contains preload directive",
      "message": "Header requirement error: For preload list removal, the header must not contain the `preload` directive."
    }
  ],
  "warnings": []
}
```

lgarron:
> Wow, it's been 8 years, and this still hasn't been merged?

That's mostly up to the maintainers with merge access, primarily @nharper.

Issue #143, opened 2018-03-01
Misleading results for subdomains and preloaded TLDs


ericlaw1979:
Repro: 
1. Visit hstspreload.org
2. Enter blog.google and hit Enter

Expect:
This domain is preloaded

Actual:
Claims it's not preloaded.

This was recently fixed in the HSTSPreload API, but that change doesn't impact the website. Discussion of this limitation is here: https://github.com/chromium/hstspreload/pull/102#issuecomment-353184441

ericlaw1979:
The same issue can be seen if you query a subdomain of a preloaded eTLD+1, e.g. `Status: www.bayden.com is not preloaded.` despite the fact that "bayden.com" is preloaded.

ericlaw1979:
As a hacky workaround, I suppose we could just issue more queries; dropping one leading TLD label each time.

lgarron:
https://github.com/chromium/hstspreload.org/issues/52 / https://github.com/chromium/hstspreload/issues/85 ?

ericlaw1979:
The change to hstspreload didn't seem to impact hstspreload.org. I haven't yet dug into why that is.

jianglai:
The change to hstspreload API doesn't fix the problem on hstspreload.org because the website does not use the API :-D

As I stated on https://github.com/chromium/hstspreload/pull/102#issuecomment-353184441, the website uses its own cache of domain states stored in datastore, supposedly to avoid loading the preload list in each request.

ExE-Boss:
This has now been fixed by #145

PR #144, opened 2018-03-13
Store IncludeSubDomains information in datastore entities

### LINKED ISSUE #52: API returns preloaded "No" for subdomains where parent domain is preloaded
If I do a query for gist.github.com, it will say that it's not preloaded, even though the parent domain (github.com) is preloaded.

This is a bit confusing because the HTTP Observatory says that it's pinned (which it is, because includeSubDomains is set with preload), but the hstspreload results say that it isn't.

Also, I have just some general curmudgeonly issues about the response being "Yes" or "No", instead of a boolean.  That said, a great option would be "Yes", "No", and "viaParentDomain" or something like that.

jianglai:
This is the first step to fix #52 (#143). Namely that when querying the preload status of an website on hstspreload.org, it does not take into account if any of the domain's ancestor domains is preloaded and includes subdomains.

For example, `.google` is preloaded and includes subdomains, but when querying for `blog.google`, hstspreload.org still returns false.

hstspreload.org app stores domain status in datastore and updates it by calling `/api/v2/update` to sync the cache with the actual preload list. Currently the datastore entities do not include information on if the domain includes subdomains. This PR modified `/api/v2/update` to pull in and store that information as a boolean property of the entity.

The updated function still relies on domain names along to determine the set of domains to be updated, removed or rejected, so this part of the logic should be consistent going forward. However that means that calling this method alone will not update the `IncludeSubDomains` property on existing entities if there were no other changes otherwise to be applied to them. A separate, temporary handler (on `/api/v2/update-includesubdomains`) is added, whose sole purpose is to update that property on existing entities based on the preload list. It should be used only when migrating data to the new schema. After all data has been updated, this method, along with its unit tests, will be deleted in the next PR.

The next PR will update `/api/v2/status` handler and implement logic to recursively look up the domain name chain to look for preloaded ancestor domains that include subdomains.

I deployed the app on a test project and tested the updated methods, all seem to be working as expected.


jianglai:
@april FYI

lgarron:
Thanks for submitting this! The overall approach looks reasonable. :-)

I was initially concerned about the performance implications of replacing `DomainsWithStatus` -> `StatesWithStatus`, but your implementation cleverly avoids that. It does split up the database query into two steps between which the database may change, but that's not worse than other places in the code base (and doesn't have severe consequences, I believe).

I'll comb through both of your PRs in detail this week.
@ericlaw1979: Do you also want to review them in detail?

ericlaw1979:
This LGTM. Since @nharper will be the one to deploy the site, I'll let him merge (and review if he likes). He'll also be the one to invoke the `update-includesubdomains` method after he deploys.

jianglai:
Friendly ping. @nharper can you take a look?

Thanks!

database/database.go
```diff
@@ -181,7 +181,25 @@ func (db DatastoreBacked) AllDomainStates() (states []DomainState, err error) {
 	return db.statesForQuery(datastore.NewQuery("DomainState"))
 }
 
-// DomainsWithStatus returns the domains with the given status in the database.
-func (db DatastoreBacked) DomainsWithStatus(status PreloadStatus) (domains []string, err error) {
-	return db.domainsForQuery(datastore.NewQuery("DomainState").Filter("Status =", string(status)))
+// StatesWithStatus returns the states of domains with the given status in the database.
+func (db DatastoreBacked) StatesWithStatus(status PreloadStatus) (domains []DomainState, err error) {
+	// Calling StatesForQuery results in timeout. Instead, call domainsForQuery, which returns keys
```
nharper: nit: statesForQuery

jianglai: Good catch!

jianglai:
PTAL. Thanks for the review!

PR #145, opened 2018-03-13
Display correct preload status for subdomains of preloaded domains

### LINKED ISSUE #52: API returns preloaded "No" for subdomains where parent domain is preloaded

jianglai:
This is part 2 to fix #52 (#143). Part one is #144. This PR is rebased on #144 so it includes all its changes. I will rebase off master once #144 is merged.

The handler of `/api/v2/status` is updated to respect preload by virtue of preloaded ancestor domains. If a given domain's status is `unknown`, we walk up the domain name chain, until we find a ancestor domain that is preloaded, and has `IncludeSubDomains = true`. Otherwise `unknown` is returned.

The query for domain status in this handler is now a cached version of `database.DomainForState`. The reason being that when walking up the chain, the TLDs are queried many more times than before. Any query to `<somedomain>.com` will also query for `com`'s state in datastore if that exact domain does not exist. Using cached queries protects us from hitting hot keys (`com` in this example) in datastore and degrading its performance. The consistency issue is minor as the preload list is not updated frequently.

Note that `IncludeSubDomains` only applies to `preloaded` status. If `www.mydomain.com`'s status is `unknown`, but `mydomain.com`'s status is `removed` and `IncludeSubDomains = true`, we do *not* consider `www.mydomain.com`'s status as `removed` and continue walking up the chain to look for a match.

`/api/v2/update-includesubdomains` and its related handler functions introduced in #144 is removed as they are only used during data migration (which turned out to be not necessary, `/api/v2/update` is able to update existing datastore entities. The newly introduced `StatesWithStatus` function combines two queries, one with `IncludeSubDomains = true` and one with `IncludeSubDomains = false`. For entities that lack that field, they are not included in either queries. As a result when reading existing entities from the datastore, none were read, and all entities from the preload list are considered new entries. These new entries are then all written into datastore, with the field included, effectively updating all existing entities).

Lastly, the unit tests in api_test.go seem to use system clock, not a fake clock. Doesn't that make these tests flaky as you cannot control how the clock advances, and therefore the cache expiry?


jianglai:
@april FYI

jianglai:
Removed `/update-includesubdomains` and updated the comment. PTAL @nharper.

jianglai:
Friendly ping @nharper.

jianglai:
Done. PTAL.

PR #156, opened 2018-07-01
Remove DoS error handler. It never worked.

lgarron:
I tried using it to mitigate an attack, but it turns out that App Engine DoS protection doesn't work for Flex / Managed Environment.

The handler also doesn't do anything useful anymore, since the error file is already gone from the repo.


lgarron:
Thanks!

PR #160, opened 2018-08-26
trim() input for domain

injust:
Leading and trailing whitespace is not currently removed, and is turned into `+` when the hostname is extracted.
We should `.trim()` the input before `extractDomain()` is called.


XhmikosR:
This does seem like a good change. Does this change apply clean @nharper? Also, does it cover all inputs?

PR #167, opened 2019-01-23
Update to use Go 1.10

nharper:


lgarron:
*woo*
Does this mean the prod server can be updated to 1.9? :-D

nharper:
I'm working on checking that. https://cloud.google.com/appengine/docs/go/ says the flex environment supports Go 1.11, 1.10, and 1.9, while https://cloud.google.com/appengine/docs/flexible/go/reference/app-yaml#general only lists 1.10, 1.9, and 1.8. I'm trying to update it to 1.10 now.

PR #173, opened 2019-03-24
Add link to actual list

hannob:
I think it would be nice if the webpage for the HSTS preload list contained info where to actually get the preload list.

The link goes to the github mirror of the chromium source, because the original on googlesource does not support direct downloading of the raw file.

### DIFF
--- a/frontend/index.html
+++ b/frontend/index.html
@@ -226,6 +226,13 @@
     </p>
   </section>
 
+  <section id="list">
+    <h2><a class="hash-link" href="#list">Getting the List</a></h2>
+    <p>
+      The HSTS preload list is part of the Chromium source code. You can download <a href="https://github.com/chromium/chromium/raw/master/net/http/transport_security_state_static.json">the HSTS preload list in JSON format here</a>.
+    </p>
+  </section>
+
   <section id="contact">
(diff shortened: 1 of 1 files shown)


lgarron:
What would you say the main reasons are that people need access to the full list? As with the public suffix list, there are issues with encouraging general consumption of the list. And most visitors to hstspreload.org are concerned about individual domains.

> The link goes to the github mirror of the chromium source, because the original on googlesource does not support direct downloading of the raw file.

I don't think it's a good idea to link to the mirror, since it's not the canonical source.

hannob:
There is no canonical source of direct download of that list, the mirror is the only one. I'd say this is an unfortunate limitation of googlesource, but I guess that's a separate issue.

As for the reason people might want to download the list I can only speak for myself, but I regularly want to check whether domains are in that list for research purposes, e.g. I might want to verify what security measures a company/service uses and just grep their domain in the list to see if they use preloading. Of course I can just put the link into my bookmarks, but I feel given there's an official page of the preload list not providing an easy way to get the actual list seems unusual.

spaze:
FWIW, the list in JSON can now be downloaded directly from cs.chromium.org: https://cs.chromium.org/codesearch/f/chromium/src/net/http/transport_security_state_static.json

PR #178, opened 2019-07-02
Upgrade to Go 1.12 and enable TLS 1.3

nharper:


nharper:
One of the annoying bits is https://cloud.google.com/appengine/docs/flexible/go/reference/app-yaml#general lists supported release versions for the runtime as go1.10, go1.9, or go1.8 (no mention of the go1.12 which appears to work).

TLS 1.3 sites will get an obsolete cipher suite warning until https://github.com/nharper/hstspreload/commit/179f6bc8a56e0541265a74905934e5da871b7af6 is landed; that's currently blocked on figuring out tests.

PR #188, opened 2020-01-27
Update index.html

XhmikosR:
Fix favicon path


nharper:
Thanks for noticing that this is broken. How about changing the paths to be absolute (with implicit domain name) instead of relative (in both index.html files)? That could be done for other paths as well.

XhmikosR:
TBH I was about to suggest this. I always use paths relative to the server root.

But maybe this should be done in another PR?

nharper:
Thanks for fixing this in #190.

Issue #159, opened 2018-08-10
Removal form displays "Error: Domain protected" for subdomains of bulk-preloaded domains


nharper:
The "Error: Domain protected" message shouldn't appear for subdomains of a bulk-preloaded domain in the removal form, and instead a separate error should appear indicating that the preloaded domain (and not a subdomain) should be entered into the removal form.

PR #196, opened 2020-08-28
Fix error messages when attempting to remove a domain not on the list.

nharper:
If one enters a domain in the removal form that is not on the preload
list, but it is preloaded because of a parent domain (or whole TLD
that is preloaded), the form states that the domain is protected from
removal and one should email hstspreload@chromium.org for removal
instead. It is more useful to display that a subdomain was entered and
the user should try again with the parent domain.

### DIFF
--- a/frontend/static/js/form.js
+++ b/frontend/static/js/form.js
@@ -268,6 +268,22 @@ PreloadController.prototype = {
   }
 };
 
+function issuesForSubdomainRemoval(status) {
+  var parentDomain = status.preloadedDomain;
+  var parentIsTLD = parentDomain.indexOf('.') == -1;
+  var error = {};
+  if (parentIsTLD) {
+    error.summary = "Domain is registered under a preloaded TLD";
+    error.message = "The entire TLD " + parentDomain + " is preloaded for HSTS and individual domain names cannot be removed.";
+  } else {
+    error.summary = "Domain is subdomain of a preloaded domain";
+    error.message = "This domain is a subdomain domain of " + parentDomain +
+                    ", which is on the preload list. To remove the HSTS " +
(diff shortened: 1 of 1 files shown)


nharper:
This PR depends on #195; it should fix issue #159

frontend/static/js/form.js
```diff
@@ -268,6 +268,22 @@ PreloadController.prototype = {
   }
 };
 
+function issuesForSubdomainRemoval(status) {
+  var parentDomain = status.preloadedDomain;
```
lgarron, line 272: Nit: `ancestorDomain` might be more accurate, since this might be a grandparent, etc.

frontend/static/js/form.js
```diff
@@ -268,6 +268,22 @@ PreloadController.prototype = {
...
+  var parentIsTLD = parentDomain.indexOf('.') == -1;
+  var error = {};
+  if (parentIsTLD) {
+    error.summary = "Domain is registered under a preloaded TLD";
+    error.message = "The entire TLD " + parentDomain + " is preloaded for HSTS and individual domain names cannot be removed.";
+  } else {
+    error.summary = "Domain is subdomain of a preloaded domain";
+    error.message = "This domain is a subdomain domain of " + parentDomain +
+                    ", which is on the preload list. To remove the HSTS " +
+                    "policy for " + status.name + ", you must remove " +
```
lgarron, line 282: Perhaps ..."example.com would need to be removed from the preload list" to avoid implying that the subdomain owner has authority over the ancestor?

nharper:
I agree that doing it on the server would be better. I'm working on a separate PR to do that.

nharper:
Abandoning this PR in favor of #197.

PR #197, opened 2020-09-10
Improve errors when attempting to remove a domain not on the list

### LINKED ISSUE #159: Removal form displays "Error: Domain protected" for subdomains of bulk-preloaded domains

nharper:
If a user enters a domain not on the preload list into the removal form,
but it is a subdomain of a preloaded domain, the error message will now
indicate that the entered domain name is a subdomain of a preloaded
domain instead of stating that the entered domain is protected from removal.

Fixes #159

### DIFF
--- a/api/api_test.go
+++ b/api/api_test.go
@@ -33,6 +33,18 @@ var issuesRemovableProtected = hstspreload.Issues{
 	},
 }
 
+var issuesRemovableSubdomain = hstspreload.Issues{
+	Errors: []hstspreload.Issue{
+		{Code: "server.removable.subdomain"},
+	},
+}
+
+var issuesRemovablePreloadedTld = hstspreload.Issues{
+	Errors: []hstspreload.Issue{
+		{Code: "server.removable.preloaded_tld"},
+	},
+}
+
 var issuesRemoveProtected = hstspreload.Issues{
 	Errors: []hstspreload.Issue{
 		{Code: "server.remove.protected"},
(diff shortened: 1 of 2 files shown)


ericroman920 (approved):

api/api_test.go
```diff
@@ -33,6 +33,18 @@ var issuesRemovableProtected = hstspreload.Issues{
 	},
 }
 
+var issuesRemovableSubdomain = hstspreload.Issues{
+	Errors: []hstspreload.Issue{
+		{Code: "server.removable.subdomain"},
```
lgarron, line 38: Nit: we use `server.remove.*` already.
```suggestion
		{Code: "server.remove.subdomain"},
```

api/api_test.go
```diff
@@ -33,6 +33,18 @@ var issuesRemovableProtected = hstspreload.Issues{
...
+	},
+}
+
+var issuesRemovablePreloadedTld = hstspreload.Issues{
+	Errors: []hstspreload.Issue{
+		{Code: "server.removable.preloaded_tld"},
```
lgarron, line 44: Nit: we use `server.remove.*` already.
```suggestion
		{Code: "server.remove.preloaded_tld"},
```

lgarron (approved):
I have some optional nits about error code consistency, but these messages will be very useful!

api/api_test.go
nharper: I think that `server.removable` is correct for these, as they are error codes for the Removable API endpoint, not the Remove API endpoint. (E.g. there's both a `server.removable.protected` and a `server.remove.protected` error code, depending on whether the request was made to /api/v2/remove or /api/v2/removable.)

PR #201, opened 2020-11-27
Add fork-corner

WarenGonzaga:
I add the modern and easiest way to place a [fork corner](https://github.com/warengonzaga/fork-corner) to the homepage.

### DIFF
--- a/frontend/index.html
+++ b/frontend/index.html
@@ -11,6 +11,11 @@
   <link rel="stylesheet" href="/static/css/style.css">
   <link rel="stylesheet" href="/static/css/form.css">
   <link rel="stylesheet" href="/static/css/github.css">
+
+  <!-- Fork Corner CDN -->
+  <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/fork-corner/dist/fork-corner.min.css">
+  <script src="https://cdn.jsdelivr.net/npm/fork-corner/dist/fork-corner.min.js" defer></script>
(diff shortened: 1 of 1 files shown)


WarenGonzaga:
@googlebot I fixed it.

WarenGonzaga:
@googlebot I fixed it.

WarenGonzaga:
@googlebot I fixed it.

lgarron:
This site is an important piece of web security infrastructure and very internationally doesn't include any third-party content.

I appreciate the problem you're trying to solve, but your project looks fairly new and this project already has a working self-contained (vendored) GitHub ribbon that has reasonably small single-file CSS implementation with accessible fallback text.

The decision is up to @nharper but in this case I would advise to kept current approach.

WarenGonzaga:
> This site is an important piece of web security infrastructure and very internationally doesn't include any third-party content.
> 
> I appreciate the problem you're trying to solve, but your project looks fairly new and this project already has a working self-contained (vendored) GitHub ribbon that has reasonably small single-file CSS implementation with accessible fallback text.
> 
> The decision is up to @nharper but in this case I would advise to kept current approach.

I understand, thanks for the input...

nharper:
I agree with @lgarron that it makes sense to keep the ribbon as is.

WarenGonzaga:
Thanks @nharper, I'm gonna close this PR.

PR #204, opened 2021-12-23
Fix grammar on success message

bdunavant:


bdunavant:
I have signed the CLA now if you want to re-scan.

lgarron (approved):

bdunavant:
According to the docs, an administraor of this repo needs to trigger a rescan of the cla/google (click on details, and it should be on the left).   I have double checked, and it says I'm set for the group, so it should work 👍

lgarron:
> According to the docs, an administraor of this repo needs to trigger a rescan of the cla/google (click on details, and it should be on the left). I have double checked, and it says I'm set for the group, so it should work 👍

Done!

(Although I can't deploy, so I don't want to merge.)

PR #207, opened 2022-04-09
Convert scripts to Python3

christhompson:
This updates the Python scripts in `scripts/` to be Python3 compatible, and does some general cleanup. This also adds a new `.pylintrc` file and `.style.yapf` file so pylint and yapf will work and apply the appropriate style guide ("Google Python Style") to make it easier to check style issues in the future. The biggest style change is moving away from printf-style format strings to Python `f""` format strings, and changing one "tricky" way of convincing the JSON parser to handle an object to a slightly different style (and adding a comment).

I've tested the updated scripts by manually running `update_bulk_preloaded.py` and by running through the playbook at https://github.com/chromium/hstspreload.org/wiki/Preload-List-Processes#updating-bulk-hsts-entries to test `roll_preload_list.py`.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -1,0 +1,441 @@
+# This Pylint rcfile contains a best-effort configuration to uphold the
+# best-practices and style described in the Google Python style guide:
+#   https://google.github.io/styleguide/pyguide.html
+#
+# Its canonical open-source location is:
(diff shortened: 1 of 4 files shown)


lgarron (approved):
The Python changes look like they don't affect functionality, thanks for keeping this project modern. :-D

PR #219, opened 2023-06-06
Update wording on site

christhompson:
Adds reminders in a couple places for site owners to also consider non-publicly accessible internal subdomains (a common pain point for preloading). Removes a blurb about preloading being a "best practice" for frameworks to consider.

### DIFF
--- a/frontend/index.html
+++ b/frontend/index.html
@@ -111,6 +111,7 @@
       <li>Serve all <strong>subdomains</strong> over HTTPS.
         <ul>
           <li>In particular, you must support HTTPS for the <code>www</code> subdomain if a DNS record for that subdomain exists.</li>
+          <li><strong>Note:</strong> HSTS preloading applies to <em>all</em> subdomains, including internal subdomains that are not publicly accessible.</li>
         </ul>
       </li>
       <li>Serve an <strong>HSTS header</strong> on the base domain for HTTPS requests:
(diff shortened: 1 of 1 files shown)


frontend/index.html
```diff
@@ -201,7 +206,7 @@ <h2><a class="hash-link" href="#opt-in">Preloading Should Be Opt-In</a></h2>
...
     </p>
     <p>
-      It's great to support HSTS preloading as a best practice, and for projects to provide a simple option to enable it. However, site operators who enable HSTS should know about the long-term consequences of preloading before they turn it on for a given domain. They should also be informed that they need to meet additional requirements and submit their site to <a href="https://hstspreload.org/">hstspreload.org</a> to ensure that it is successfully preloaded (i.e. to get the full protection of the intended configuration).
+      Site operators who enable HSTS should know about the long-term consequences of preloading before they turn it on for a given domain. They should also be informed that they need to meet additional requirements and submit their site to <a href="https://hstspreload.org/">hstspreload.org</a> to ensure that it is successfully preloaded (i.e. to get the full protection of the intended configuration).
```
nharper, line 209: The removal of the first sentence of this paragraph looks good. In the edit, what's now the first sentence looks like it's addressed at a different audience (site operators) than this section is addressed at (projects that provide HTTPS configuration). I think the gist of this paragraph is that projects configuring HTTPS need informed consent from the site operator before enabling HSTS or HSTS preload. Is there an easy way to edit this to better align with that goal?

nharper (commented):
Most of this looks good. One comment on possibly improving one section.

frontend/index.html
christhompson: Re-wrote this a bit to make the subject be about such projects (rather than directed at site owners). Let me know what you think.

nharper (approved):

christhompson:
Thanks! Looks like I'm not a repository member so I can't merge this myself.

PR #221, opened 2023-06-13
GatherList function addition

tKabiawu:
Created a function that takes a preload list as input and returns two string slices of domain names. The first slice is for domains that were submitted to the preload list with policy "bulk-18-weeks" and the second slice is for domains that were submitted to the list with policy "bulk-1-year".


api/gather.go
```diff
@@ -0,0 +1,22 @@
+package api
+
+import (
+	"github.com/chromium/hstspreload/chromium/preloadlist"
+)
+
+// gatherLists returns a list of domains filtered by Policy types "bulk-18-weeks" and "bulk-1-year"
+func gatherLists(list preloadlist.PreloadList) ([]string, []string) {
```
lgarron, line 8: This is vague and ambiguous in several ways:

- The return types are not annotated. It's very easy to mis-use them. I would suggest using a map or a struct.
- The function signature (and comment) do not indicate that this drops entries.

Also:

- This function is not capitalized, so it's not exported. But it's also not used internally. So it's not doing anything.
- Since the strings are constants, it would be good to define them as capitalized constants, ideally as an anum.
- Any reason to return just the domain strings rather than the entries?
- Any reason to implement this in the `hstspreload.org` package rather than `hstspreload`?

I also don't see a stated motivation for this PR. If the goal is to be able to get subsets of the list, would it make just as much sense to provide filters for individual states rather than gathering some subsets at once?

tKabiawu: Hi Lucas,

Thanks so much for your comments they were very helpful. I'm currently a step intern working under @nharper and @carlosjoan91, and the project we are working on revolves around automating processes concerning the hsts preload list. I have attempted to make changes to the code that adheres to some of your suggestions by provide functions that filter for individual states rather than gather some subsets at once, returning entries instead of  strings (thereby annotating the return type), capitalizing the function so that it is exportable, and defining constants as enums. Initially, I chose to return just the domain strings rather than the entries because I initially thought we only needed the domain name to check whether the domain is preloadable with the PreloadableDomain() function. However, I forgot that to change the status of a domain, you also need the value of the includeSubDomains field, which led me to choose to return the entire entry. Moreover, my understanding as the reason this is implemented in the ```hstspreload.org ```package instead of the ```hstspreload``` is because we plan to use a task scheduling tool that will call the endpoint at regular intervals.

lgarron: Ah, thanks, I was missing the context! Happy to see work on this project, but I'm glad to leave you in @nharper and @carlosjoan91's capable hands for an actual review. 😄

nharper (commented):
The `api` directory doesn't feel like the right place for this to live. I'm not sure where the right place is, but I think it should live alongside or nearby its callers.

api/gather.go
```diff
@@ -0,0 +1,42 @@
...
+type PolicyType int
+
+const (
+	bulk18Week PolicyType = iota
+	bulk1Year
+)
+
+func (pt PolicyType) String() string {
+	return []string{"bulk-18-weeks", "bulk-1-year"}[pt]
+}
```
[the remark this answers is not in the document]
tKabiawu: Done

PR #223, opened 2023-06-14
datastore interface for automated removals

bkhushi:


bkhushi:
@nharper @carlosjoan91 could you please review this pr?

nharper (commented):
Overall the structure here looks good. The tests could use some work.

database/database.go
```diff
@@ -27,6 +28,9 @@ type Database interface {
 	StateForDomain(string) (DomainState, error)
 	AllDomainStates() ([]DomainState, error)
 	StatesWithStatus(PreloadStatus) ([]DomainState, error)
+	GetIneligibleDomains(string) (IneligibleDomainState, error)
```
carlosjoan91, line 31: Looks like this needs to be []string in the parameters

database/database.go
```diff
@@ -186,3 +190,111 @@ func (db DatastoreBacked) StatesWithStatus(status PreloadStatus) (domains []Doma
 	return db.statesForQuery(
 		datastore.NewQuery("DomainState").Filter("Status =", string(status)))
 }
+
+// GetIneligibleDomain returns the state for the given domain.
+func (db DatastoreBacked) GetIneligibleDomains(domains []string) (states []IneligibleDomainState, err error) {
```
carlosjoan91, line 195: The name of the function makes it sound like this returns a list of domains, but in reality this returns the states for the domains it is given, I think GetIneligibleDomainsState or something like that would probably be more descriptive.

database/database_test.go
```diff
@@ -234,3 +234,180 @@ func TestStatesWithStatus(t *testing.T) {
...
+	},
+	{
+		"bulk-1-year",
+		[]IneligibleDomainState{{Name: "wikipedia.com", Policy: "bulk-1-year"}},
+		[]string{"Updating 1 entries...", " done.\n"},
+	},
+}
+
+// Test SetIneligibleDomainStates
+func TestSet(t *testing.T) {
```
carlosjoan91, line 265: nit: Here and in other functions, it's common to name test functions TestNameOfFunctionTested, so this would become TestSetInelegibleDomains

database/database_test.go
```diff
@@ -234,3 +234,180 @@ func TestStatesWithStatus(t *testing.T) {
...
+			{Name: "g.com", Policy: "bulk-18-week"},
+			{Name: "h.com", Policy: "bulk-18-week"},
+			{Name: "i.com", Policy: "bulk-1-year"},
+			{Name: "j.com", Policy: "bulk-18-week"},
+			{Name: "k.com", Policy: "bulk-1-year"},
+		},
+	},
+}
+
+// Test Set and Get
```
carlosjoan91, line 326: This comment might be outdated since Set now has its own test

database/database_test.go
```diff
@@ -234,3 +234,180 @@ func TestStatesWithStatus(t *testing.T) {
...
+		}
+		for i, domainState := range domainStates {
+			if domainState.Name != (tr.domainStates[i]) {
+				t.Errorf("unexpected domain at position %d for test %s: %#v", i, tr.description, domainState)
+			}
+		}
+	}
+}
+
+// TestDelete tests the DeleteIneligibleDomains
```
carlosjoan91, line 373: nit: ... the DeleteInelegibleDomains function.

database/ineligibledomainstate.go
```diff
@@ -0,0 +1,27 @@
...
+)
+
+// IneligibleDomainState contains the state about a domain name that 
+// is potentially ineligible to remain on the list and is at risk 
+// for being removed from the list
+type IneligibleDomainState struct {
+	// Name is the key in the datastore, so we don't include it as a field
+	// in the stored value.
+	Name string `datastore:"-" json:"name"`
+	// Scan is where information of the checks are stored
```
carlosjoan91, line 16: Nit: Scans

carlosjoan91 (commented):
Thanks! Left a few comments, mostly about code style

database/database_test.go
```diff
@@ -234,3 +234,180 @@ func TestStatesWithStatus(t *testing.T) {
...
+var getAndDeleteTests = []struct {
+	description  string
+	domainStates []string
+	wantStates   []IneligibleDomainState
+}{
+	{
+		"one domain",
+		[]string{"a.com"},
+		[]IneligibleDomainState{
+			{Name: "a.com", Policy: "bulk-1-year"},
```
lgarron, line 302: Total drive-by-nit, but I would recommend against using any test domains names that are registerable, even if the real domains are not meant accessed in tests. (In fact, the HSTS preload list itself has caused issues for people who assumed it was safe to use test domains that could and did actually become registerable.)

`.test` and `.example` are safe choices: https://www.rfc-editor.org/rfc/rfc2606.html#section-2

bkhushi: This is very interesting and something I hadn't thought about. I went ahead and fixed it! Thank you for bringing this up

database/database_test.go
```diff
@@ -62,46 +62,46 @@ var putAndAllTests = []struct {
 	{
 		"one domain",
 		[]DomainState{
-			{Name: "gmail.com", Status: StatusPending},
+			{Name: "gmail.test", Status: StatusPending},
```
nharper, line 65: nit: it would be better for the renames in lines otherwise untouched by this PR to be in a separate PR that just does cleanup work. (You don't have to do the cleanup of existing code.)

database/database_test.go
```diff
@@ -234,3 +234,180 @@ func TestStatesWithStatus(t *testing.T) {
 		}
 	}
 }
+
+// setTests is a struct that is used in testing SetIneligibleDomainStates
+var setTests = []struct {
```
nharper, line 239: The name `setTests` is too generic ("set" is a common db operation, while this is specific to IneligibleDomainStates). Either rename it to be more specific, or move it inside of TestSetIneligibleDomainStates.

database/database_test.go
```diff
@@ -234,3 +234,180 @@ func TestStatesWithStatus(t *testing.T) {
...
+		[]string{"Updating 1 entries...", " done.\n"},
+	},
+	{
+		"bulk-1-year",
+		[]IneligibleDomainState{{Name: "wikipedia.test", Policy: "bulk-1-year"}},
+		[]string{"Updating 1 entries...", " done.\n"},
+	},
+}
+
+// Test SetIneligibleDomainStates
```
nharper, line 264: nit: elaborate in the comment what aspects of SetIneligibleDomainStates this is testing

database/database_test.go
```diff
@@ -234,3 +234,180 @@ func TestStatesWithStatus(t *testing.T) {
...
+			t.Errorf("[%s] Incorrect status reports: %#v", tt.description, statuses)
+		}
+	}
+}
+
+// getAndDeleteTests is a struct that is used in testing GetIneligibleDomainStates
+// and DeleteIneligibleDomainStates
+var getAndDeleteTests = []struct {
+	description  string
+	domainStates []string
```
nharper, line 295: I'd expect a field with a name `domainStates` to have a type like `[]DomainState` or `[]IneligibleDomainState`. Perhaps this should be `domains` or `domainNames` instead?

database/database_test.go
```diff
@@ -234,3 +234,180 @@ func TestStatesWithStatus(t *testing.T) {
...
+			{Name: "e.test", Policy: "bulk-1-year"},
+			{Name: "f.test", Policy: "bulk-1-year"},
+			{Name: "g.test", Policy: "bulk-18-week"},
+			{Name: "h.test", Policy: "bulk-18-week"},
+			{Name: "i.test", Policy: "bulk-1-year"},
+			{Name: "j.test", Policy: "bulk-18-week"},
+			{Name: "k.test", Policy: "bulk-1-year"},
+		},
+	},
+}
```
nharper, line 324: The IneligibleDomainState literals here all only have Name and Policy filled out. I would think setting and getting the Scans field would also be important and it should be in most of the test cases.

database/database_test.go
```diff
@@ -234,3 +234,180 @@ func TestStatesWithStatus(t *testing.T) {
...
+		if err != nil {
+			t.Errorf("%s", err)
+		}
+	}
+
+	if len(domainStates) != 0 {
+		t.Errorf("Empty database should contain no preloaded domains")
+	}
+
+}
```
nharper, line 413: Some thoughts of other things to test:

- calls to the new functions with duplicate entries
- calls to get and delete for entries that don't exist

database/database_test.go
```diff
@@ -234,3 +234,180 @@ func TestStatesWithStatus(t *testing.T) {
...
+// Test GetIneligibleDomainStates
+func TestGetIneligibleDomainStates(t *testing.T) {
+
+	resetDB()
+
+	// domainStates should be empty as domains are not added to database
+	domainStates, err := testDB.GetIneligibleDomainStates([]string{"a.test"})
+	if len(domainStates) != 0 {
+		t.Errorf("Empty database should contain no preloaded domains")
+	}
```
nharper, line 335: If the intention here is to check that the database is empty, it's fine to use `testDB.backend` directly to get all IneligibleDomainStates (in a helper function that can be used here and in TestDeleteIneligibleDomainStates). It's also fine to assume that `resetDB` has done just that and the database is empty.

database/database_test.go
bkhushi: I added a new function that tests for duplicate entries.  This would also test delete for entries that do not exist and get entries that do not exist is tested on line 462!

database/database_test.go
```diff
@@ -234,3 +236,423 @@ func TestStatesWithStatus(t *testing.T) {
...
+		sort.Slice(domainStates, func(i, j int) bool { return domainStates[i].Name < domainStates[j].Name })
+		if len(domainStates) != len(tr.domainNames) {
+			t.Errorf("Incorrect count of states for test %s", tr.description)
+		}
+		for i, domainState := range domainStates {
+			if domainState.Name != (tr.domainNames[i]) {
+				t.Errorf("unexpected domain at position %d for test %s: %#v", i, tr.description, domainState)
+			}
+		}
+	}
```
nharper, line 450: table driven tests should do only one loop over the table. Treat the test like you're writing a test for only one entry in the table.

database/database_test.go
```diff
@@ -234,3 +236,423 @@ func TestStatesWithStatus(t *testing.T) {
...
+// database
+func TestGetIneligibleDomainStates(t *testing.T) {
+
+	resetDB()
+
+	// domainStates should be empty as domains are not added to database
+	// test for entry that does not exist
+	domainStates, err := testDB.GetIneligibleDomainStates([]string{"a.test"})
+	if len(domainStates) != 0 {
+		t.Errorf("Empty database should contain no preloaded domains")
```
nharper, line 414: This string looks like it was copied/pasted, as it refers to preloaded domains instead of ineligible domain states.

database/database_test.go
```diff
@@ -234,3 +236,423 @@ func TestStatesWithStatus(t *testing.T) {
...
+	for _, tr := range getAndDeleteTests {
+		err = testDB.DeleteIneligibleDomainStates(tr.domainNames)
+		if err != nil {
+			t.Errorf("%s", err)
+		}
+	}
+
+	if len(domainStates) != 0 {
+		t.Errorf("Empty database should contain no preloaded domains")
+	}
```
nharper, line 491: `domainStates` is unchanged from line 459, so this is redundant (it was checked on line 461). I think what you want to do here is check that the database contains no IneligibleDomainStates after the delete operation.

database/database_test.go
```diff
@@ -234,3 +236,423 @@ func TestStatesWithStatus(t *testing.T) {
...
+		for i, domainState := range domainStates {
+			if domainState.Name != (tr.domainNames[i]) {
+				t.Errorf("unexpected domain at position %d for test %s: %#v", i, tr.description, domainState)
+			}
+		}
+	}
+}
+
+// TestDeleteIneligibleDomainStates tests the DeleteIneligibleDomainStates function
+func TestDeleteIneligibleDomainStates(t *testing.T) {
```
nharper, line 454: All of the comments on TestGetIneligibleDomainStates apply here as well.

database/database_test.go
```diff
@@ -234,3 +236,423 @@ func TestStatesWithStatus(t *testing.T) {
...
+	}
+
+	if len(domainStates) != 0 {
+		t.Errorf("Empty database should contain no preloaded domains")
+	}
+
+}
+
+// setIneligibleDomainTests is a struct that is used in testing SetIneligibleDomainStates
+var setGetDeleteDuplicateIneligibleDomainTests = []struct {
```
nharper, line 496: I don't understand the difference between the two test table entries here. The test would be much easier to understand with only one (and thus no table).

database/database_test.go
```diff
@@ -234,3 +236,423 @@ func TestStatesWithStatus(t *testing.T) {
...
+				},
+			},
+		}},},
+		[]string{"done. \n"},
+	},
+}
+
+// TestSetGetDeleteDuplicateIneligibleDomainStates tests adding, getting, and deleting duplicate 
+// domain states from the database
+func TestSetGetDeleteDuplicateIneligibleDomainStates(t *testing.T) {
```
nharper, line 580: This test is testing too many different things. I think it's testing some subset of the following, each of which should be a separate test:

- a call to `SetIneligibleDomainStates` with duplicate entries
- a call to `GetIneligibleDomainStates` with duplicate names
- a call to `GetIneligibleDomainStates` with a name that doesn't exist
- a call to `DeleteIneligibleDomainStates` with duplicate names
- a call to `DeleteIneligibleDomainStates` with a name that doesn't exist

database/database_test.go
```diff
@@ -234,3 +237,412 @@ func TestStatesWithStatus(t *testing.T) {
...
+					t.Errorf("%s", err)
+				}
+				_ = err
+			}
+		} else {
+			if err != datastore.ErrNoSuchEntity {
+				t.Errorf("%s", err)
+			}
+		}
+	}
```
nharper, line 643: This only needs to check that that the domain you deleted is no longer in the database.

nharper (commented):
This is looking pretty good, just a few more things to fix.

nharper (approved):

PR #224, opened 2023-06-22
Datastore Policy Field addition

tKabiawu:
This addition adds the PolicyField in the DomainState struct. We do this so that the policy underwhich domains are admitted are able to be stored in the hstspreload.org datastore when the Update API handler is hit. My reviewers are @nharper and @carlosjoan91. This satisfies b/287096462.


api/update.go
```diff
@@ -105,6 +107,7 @@ func (api API) Update(w http.ResponseWriter, r *http.Request) {
 			Message:           "Domain was added and removed without being preloaded.",
 			Status:            database.StatusRejected,
 			IncludeSubDomains: entry.IncludeSubDomains,
+			Policy:            entry.Policy,
```
nharper, line 110: I'm noticing that a lot of these look very similar (copy a bunch of fields from the `preloadlist.Entry` and set a Status). Optionally, it might be nice to have an equivalent to `DomainState.ToEntry` that goes the other way and takes in a status.

nharper (commented):
For testing, take a look at api/api_test.go. That'll generally show how we test API endpoints, though those are black box tests and here you probably want to look at more than just the HTTP response returned to check that the correct fields are written to the database. That test is still a good example of how to set up appropriate mocks, and then you could make an update_test.go file for additional testing.

api/update.go
tKabiawu: Done.

tKabiawu:
> For testing, take a look at api/api_test.go. That'll generally show how we test API endpoints, though those are black box tests and here you probably want to look at more than just the HTTP response returned to check that the correct fields are written to the database. That test is still a good example of how to set up appropriate mocks, and then you could make an update_test.go file for additional testing.

Done

api/update_test.go
```diff
@@ -0,0 +1,86 @@
...
+
+	w := httptest.NewRecorder()
+	w.Body = &bytes.Buffer{}
+
+	r, err := http.NewRequest("GET", "", nil)
+	if err != nil {
+		t.Fatalf("[%s] %s", "NewRequest Failed", err)
+	}
+
+	api.Update(w, r)
```
nharper, line 64: Check that the HTTP response is valid

api/update_test.go
```diff
@@ -0,0 +1,86 @@
...
+		{Name: "godoc.og", Mode: "", IncludeSubDomains: true, Policy: preloadlist.Custom},
+		{Name: "dev", Mode: preloadlist.ForceHTTPS, IncludeSubDomains: true, Policy: preloadlist.UnspecifiedPolicyType},
+	}}
+
+	mockHstspreload.preloadableResponses = TestPreloadableResponses
+	mockHstspreload.removableResponses = TestRemovableResponses
+	mockPreloadlist.list = TestPreloadlist
+
+	mockController.FailCalls = (failNone & failDatabase) != 0
+	mockPreloadlist.failCalls = (failNone & failChromiumpreload) != 0
```
nharper, line 54: these can be simplified

api/update_test.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+
+	api.Update(w, r)
+
+	states, err := api.database.AllDomainStates()
+	if err != nil {
+		t.Fatalf("Couldn't get the states of all domains in the database.")
+	}
+
+	expectedPolicies := map[string]preloadlist.PolicyType{
```
nharper, line 71: It looks like all of these are entries that are on the preload list. AllDomainStates will also return domains that aren't on the preload list, so you should test somewhere (maybe in a database test) that it returns the correct PolicyType for domains that aren't on the list.

Also, afaict the call to `Update` inserts all of these domains into the datastore. It would be good to test what happens when a domain is already in the datastore (e.g. is StatusPending) and then gets updated to StatusPreloaded.

database/domainstate.go
```diff
@@ -70,13 +73,23 @@ func (s DomainState) Equal(s2 DomainState) bool {
...
+}
+
+// ToDomainState converts a preloadlist.Entry to a DomainState.
+func ToDomainState(entry preloadlist.Entry, status PreloadStatus, message string) DomainState {
+	return DomainState{Name: entry.Name, Status: status, Message: message, IncludeSubDomains: entry.IncludeSubDomains, Policy: entry.Policy}
+}
+
+// ToDomainStateWithoutMessage converts a preloadlist.Entry to a DomainState with no accompanying message.
+func ToDomainStateWithoutMessage(entry preloadlist.Entry, status PreloadStatus) DomainState {
+	return ToDomainState(entry, status, "")
```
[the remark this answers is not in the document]
tKabiawu: Done.

api/update_test.go
tKabiawu: Done

PR #229, opened 2023-07-14
Removed Redeployment Requirement to track bulk preloaded domains. 

tKabiawu:
Removed update_bulk_prelaoded script and instead enabled codebase to determine a domain's bulk status by reading the Policy field for each domain in the database.


api/api_test.go
```diff
@@ -195,7 +190,7 @@ func TestAPI(t *testing.T) {
 		{"bulk status", data1, failNone, api.Status, "GET", "?domain=removal-not-preloaded-bulk-eligible.test",
 			200, jsonContentType, wantBody{bulkState: &DomainStateWithBulk{
 				DomainState: &database.DomainState{Name: "removal-not-preloaded-bulk-eligible.test", Status: database.StatusUnknown},
-				Bulk:        true,
+				Bulk:        false,
```
nharper, line 193: Given this name is removal-not-preloaded-bulk-eligible, I expect that Bulk should be true, not false. It might be that this test case doesn't make sense at all (it appears that it's testing that a domain that isn't preloaded, and has a status of StatusUnknown, should still be listed as bulk preloaded), or there might be some bug in the test setup or elsewhere in the PR. What do you think is going here, and what do you think the right fix would be?

nharper (commented):
The python script is also referenced in the Makefile. Either remove all references to it, or remove the script in a follow-up PR.

api/api_test.go
nharper: I agree with the assessment that if a domain isn't in the datastore (which matches with StatusUnknown), it has no policy, and for Bulk to be true, it needs to have a bulk policy. I.e. this test is testing a case that doesn't exist in real life.

Having the `protected()` function return a tri-state of true, false, or error feels to me like overkill. Given the definition of the function (is the domain protected from removal), it doesn't make sense to ask that question about a domain that isn't preloaded, and it would be fine to say that the function's output is undefined if the domain isn't on the preload list.

One way to avoid this ambiguity would be to get rid of the function called `protected()` and instead have a function that returns whether a domain has a bulk policy, e.g. `IsBulk`. That way, the function is well defined for all inputs.

api/domain_handlers.go
```diff
@@ -57,13 +58,19 @@ func (api API) protected(domain string, state database.PreloadStatus) bool {
 		return false
 	}
 
+	domainState, _ := api.database.StateForDomain(domain)
```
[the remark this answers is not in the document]
tKabiawu: Done

api/api_test.go
tKabiawu: Thanks, Done

nharper (commented):
One small change, otherwise this looks great

Makefile
```diff
@@ -19,10 +19,6 @@ deploy: bulk-preloaded-force-update version
 	time gcloud app deploy app.yaml
 	date
 
-.PHONY: bulk-preloaded-force-update
-bulk-preloaded-force-update:
```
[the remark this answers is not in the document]
tKabiawu: Done

nharper (approved):

PR #232, opened 2023-07-20
added EligibleDomain to hstspreloadWrapper in api/wrappers.go

bkhushi:


go.mod
```diff
@@ -5,7 +5,7 @@ go 1.15
 require (
 	cloud.google.com/go/datastore v1.11.0
 	github.com/chromium/hstspreload v0.0.0-20230623210540-9e257cb2df03
```
nharper, line 7: This is the line that needs to change (instead of the replace) so that `hstspreload.EligibleDomain` is available. `go get -u .` will update this dependency (and others).

nharper (approved):
LGTM

PR #233, opened 2023-07-21
Performs scans on domains and updates IneligibleDomainStates

bkhushi:


api/domain_handlers.go
```diff
@@ -381,3 +381,55 @@ func (api API) Remove(w http.ResponseWriter, r *http.Request) {
 
 	writeJSONOrBust(w, issues)
 }
+
+// Example: GET /ineligible?
+func (api API) Ineligible(w http.ResponseWriter, r *http.Request) {
```
nharper, line 386: Add a comment explaining what this endpoint/function does.

This endpoint/function can be renamed to be clearer about what it does.

(Eventually this needs to be registered in server.go, but it's good to leave that out of this PR, as there's additional work that we probably want to do before the server is configured to call this function.)

api/domain_handlers.go
```diff
@@ -381,3 +381,55 @@ func (api API) Remove(w http.ResponseWriter, r *http.Request) {
...
+	var deleteEligibleDomains []string
+
+	domains, err := api.database.AllDomainStates()
+	if err != nil {
+		msg := fmt.Sprintf("Internal error: could not retrieve domains. (%s)\n", err)
+		http.Error(w, msg, http.StatusInternalServerError)
+		return
+	}
+	// Store ineligible domains in slice
+	for _, d := range domains {
```
nharper, line 398: Should this be looping over all domains, or just those with bulk-18-weeks and bulk-1-year?

api/domain_handlers_test.go
```diff
@@ -24,3 +34,89 @@ func TestNormalizeDomain(t *testing.T) {
...
+		"chromium.org": emptyIssues,
+		"godoc.og":     issuesWithErrors,
+		"dev":          issuesWithWarnings,
+		"example.com":  issuesWithErrors,
+	}
+	TestPreloadlist := preloadlist.PreloadList{Entries: []preloadlist.Entry{
+		{Name: "garron.net", Mode: preloadlist.ForceHTTPS, IncludeSubDomains: true, Policy: preloadlist.Bulk18Weeks},
+		{Name: "chromium.org", Mode: preloadlist.ForceHTTPS, IncludeSubDomains: false, Policy: preloadlist.Bulk1Year},
+		{Name: "godoc.og", Mode: "", IncludeSubDomains: true, Policy: preloadlist.Bulk18Weeks},
+		{Name: "dev", Mode: preloadlist.ForceHTTPS, IncludeSubDomains: true, Policy: preloadlist.Bulk1Year},
```
nharper, line 68: this should have a policy of PublicSuffix

api/domain_handlers.go
```diff
@@ -381,3 +381,55 @@ func (api API) Remove(w http.ResponseWriter, r *http.Request) {
...
+		_, issues := hstspreload.EligibleDomain(d.Name, d.Policy)
+
+		if len(issues.Errors) > 0 {
+			ineligibleDomains = append(ineligibleDomains, database.IneligibleDomainState{
+				Name: d.Name,
+				Scans: []database.Scan{
+					{
+						ScanTime: time.Now(),
+						Issues:   []hstspreload.Issues{issues},
```
nharper, line 407: (I missed this when reviewing a previous PR.) I think that the `Issues` field in the `Scan` struct should be of type `hstspreload.Issues`, not `[]hstspreload.Issues`. The `Issues` type already encapsulates a set of errors and warnings, and as demonstrated by the usage here (the output from `hstspreload.EligibleDomain`), there will be only one `Issues` for each scan.

api/domain_handlers_test.go
```diff
@@ -24,3 +34,89 @@ func TestNormalizeDomain(t *testing.T) {
...
+		t.Errorf("HTTP Response Invalid: Status code is not 200")
+	}
+
+	states, err := api.database.GetAllIneligibleDomainStates()
+	if err != nil {
+		t.Fatalf("Couldn't get the states of all domains in the database.")
+	}
+
+	for _, state := range states {
+		if state.Scans[0].Issues[0].Match(expectedScans[state.Name].Issues[0]) {
```
nharper, line 115: A state can have more than one scan in `Scans`. We should check that all scans match the expected states.

api/domain_handlers_test.go
```diff
@@ -24,3 +34,89 @@ func TestNormalizeDomain(t *testing.T) {
...
+		Issues: []hstspreload.Issues{issuesWithErrors},
+	}
+
+	TestEligibleResponses := map[string]hstspreload.Issues{
+		"garron.net":   emptyIssues,
+		"badssl.com":   issuesWithWarnings,
+		"chromium.org": emptyIssues,
+		"godoc.og":     issuesWithErrors,
+		"dev":          issuesWithWarnings,
+		"example.com":  issuesWithErrors,
```
nharper, line 62: Instead of using real-world examples (or domain names that look close to real domains, e.g. `godoc.og`), this would be a good place to use RFC 2606 testing/documentation names (e.g. ending in `.test` or `.example`), and give them descriptive names. For example, when I see `garron.net` in this test file, I have to look at multiple places to understand what it's testing. If it's named `preloaded-bulk-18-weeks-no-issues.test`, then it's clearer what its purpose is throughout the test.

api/domain_handlers_test.go
```diff
@@ -24,3 +34,89 @@ func TestNormalizeDomain(t *testing.T) {
...
+	}
+	TestPreloadlist := preloadlist.PreloadList{Entries: []preloadlist.Entry{
+		{Name: "garron.net", Mode: preloadlist.ForceHTTPS, IncludeSubDomains: true, Policy: preloadlist.Bulk18Weeks},
+		{Name: "chromium.org", Mode: preloadlist.ForceHTTPS, IncludeSubDomains: false, Policy: preloadlist.Bulk1Year},
+		{Name: "godoc.og", Mode: "", IncludeSubDomains: true, Policy: preloadlist.Bulk18Weeks},
```
nharper, line 67: Good idea to test something with a mode other than "force-https".

The preload list (transport_security_state_static.json) has some entries that aren't for preloading HSTS. Only the entries with a mode of "force-https" are preloading HSTS. (The other entries are for preloading key pinning.) This project only cares about entries preloaded for HSTS, so for entries (like this one) that don't have a mode of "force-https", nothing should be done. The current behavior is that a scan is still done.

nharper (commented):
There are a lot of comments, but most of them should be easy/straightforward to resolve. Overall this looks pretty good.

api/domain_handlers.go
```diff
@@ -381,3 +382,70 @@ func (api API) Remove(w http.ResponseWriter, r *http.Request) {
...
+		msg := fmt.Sprintf("Internal error: could not retrieve domains. (%s)\n", err)
+		http.Error(w, msg, http.StatusInternalServerError)
+		return
+	}
+	// Store ineligible domains in slice
+	for _, d := range domains {
+		if d.Policy == preloadlist.Bulk18Weeks || d.Policy == preloadlist.Bulk1Year {
+
+			_, issues := api.hstspreload.EligibleDomain(d.Name, d.Policy)
+			state, err := api.database.GetIneligibleDomainStates([]string{d.Name})
```
nharper, line 407: Multiple repeated calls to `GetIneligibleDomainStates` isn't a good access pattern - it would be better to make one call with all domains we care about. One way to do this would be to filter the domains slice to have only the ones with policies we care about, then make this call, and then iterate over the filtered domain slice. (For ease of access, it might be convenient to make a map from domain name to IneligibleDomainState.)

Another potentially better way would be to call `GetAllIneligibleDomainStates`. The thought here is that if there's an IneligibleDomainState for a domain that's no longer on the preload list, then we should remove it. This makes logic a little more complicated in the body of this function to identify such domains. One potential way to track that is if you have a map of domains to IneligibleDomainState, you could delete from that map when you process domains, and the ones remaining in the map are the ones that need to be removed.

api/domain_handlers.go
```diff
@@ -381,3 +382,70 @@ func (api API) Remove(w http.ResponseWriter, r *http.Request) {
...
+	if err != nil {
+		msg := fmt.Sprintf("Internal error: could not retrieve domains. (%s)\n", err)
+		http.Error(w, msg, http.StatusInternalServerError)
+		return
+	}
+	// Store ineligible domains in slice
+	for _, d := range domains {
+		if d.Policy == preloadlist.Bulk18Weeks || d.Policy == preloadlist.Bulk1Year {
+
+			_, issues := api.hstspreload.EligibleDomain(d.Name, d.Policy)
```
nharper, line 406: (I missed this in the previous review comments.) This function takes some time to scan domains and return. Currently, these are done sequentially, which given the number of domains could result in this endpoint timing out. For this PR, how about adding a TODO to consider parallelizing this operation?

api/domain_handlers.go
```diff
@@ -381,3 +382,70 @@ func (api API) Remove(w http.ResponseWriter, r *http.Request) {
...
+					})
+					ineligibleDomains = append(ineligibleDomains, state[0])
+				} else {
+					ineligibleDomains = append(ineligibleDomains, database.IneligibleDomainState{
+						Name: d.Name,
+						Scans: []database.Scan{
+							{
+								ScanTime: time.Now(),
+								Issues:   issues,
+							},
```
nharper, line 422: After the call to `EligibleDomain`, you can create the `database.Scan` struct with the time instead of needing to repeat this.

api/domain_handlers_test.go
```diff
@@ -24,3 +34,94 @@ func TestNormalizeDomain(t *testing.T) {
...
+	}
+
+	TestEligibleResponses := map[string]hstspreload.Issues{
+		"preloaded-bulk-18-weeks-no-issues.test":   emptyIssues,
+		"preloaded-bulk-18-weeks-warnings.test":   issuesWithWarnings,
+		"preloaded-bulk-1-year-errors.test": issuesWithErrors,
+		"not-preloaded-bulk-18-weeks-errors.test":     issuesWithErrors,
+		"preloaded-bulk-1-year-warnings":          issuesWithWarnings,
+		"preloaded-bulk-18-weeks-errors.test":  issuesWithErrors,
+		"preloaded-public-suffix-no-issues": emptyIssues,
```
nharper, line 55: nit: align the right hand side of the map entries to be in the same column

api/domain_handlers.go
```diff
@@ -381,3 +382,94 @@ func (api API) Remove(w http.ResponseWriter, r *http.Request) {
...
+					Policy: string(d.Policy),
+				})
+			}
+		} else if ok {
+			deleteEligibleDomains = append(deleteEligibleDomains, d.Name)
+		}
+	}
+
+	// Delete eligible domains from the database
+	err = api.database.DeleteIneligibleDomainStates(deleteEligibleDomains)
```
nharper, line 460: For domains that are in `states` but aren't in `policyStates`/`policyDomains`, we should remove them from the ineligible domain state table at some point. (This could happen if a domain stops meeting the preload requirements, gets scanned and that result is marked down, and then the domain owner fully removes the domain from the preload list.)

api/domain_handlers.go
```diff
@@ -381,3 +382,101 @@ func (api API) Remove(w http.ResponseWriter, r *http.Request) {
...
+		http.Error(w, msg, http.StatusInternalServerError)
+		return
+	}
+	
+	// delete domains that exist in the ineligible database but not
+	// on the preload list
+	for _, s := range state {
+		states[s.Name] = s
+		_, ok := policyStates[s.Name]
+		if !ok {
```
nharper, line 434: nit: this can be a one-liner `if _, ok := policyStates[s.Name]; !ok {`

nharper (approved):
It looks like this needs to be rebased past PR #235, as it looks like the changes from that PR are showing up in the diff here as well.

PR #234, opened 2023-07-24
Created SetAutomatedPendingRemoval and StatesForDomains database methods

tKabiawu:
The addition of this functionality will allow for changing domain states in the database to have a status of StatusPendingAutomatedRemoval by calling the SetPendingAutoamtedRemoval function on a slice of domain name strings. The domain names must already be in the database, however.


database/database.go
```diff
@@ -153,6 +153,93 @@ func (db DatastoreBacked) StateForDomain(domain string) (state DomainState, err
...
+			domainStates[i].Name = keys[i].Name
+		}
+		return domainStates, nil
+	}
+
+	var keys []*datastore.Key
+	for _, domain := range domains {
+		key := datastore.NameKey(domainStateKind, domain, nil)
+		keys = append(keys, key)
+		if len(keys) >= batchSize {
```
carlosjoan91, line 182: I'm not sure I follow this section, this is getting the domain states for a subset of keys (< batchsize), but then not doing anything with them (unless there is an error), then at the end what we actually return is getDomainStates for all the keys. A similar approach works in the put case because each call results in an update to the DB, but since we are just reading in this case, I think this code basically is a no-op (unless I'm missing something).

database/database_test.go
```diff
@@ -177,6 +178,41 @@ func TestStateForDomain(t *testing.T) {
 	}
 }
 
+func TestStatesForDomains(t *testing.T) {
+	resetDB()
+
+	if err := testDB.PutState(DomainState{Name: "a.test", Status: StatusPreloaded}); err != nil {
+		t.Errorf("cannot put state %s for test StatatesForDomains", err)
```
carlosjoan91, line 185: nit: This and the following should instead be t.Fatalf since this is a failure in the test setup rather than a failed expectation (i.e. you are not really meaning to test the PutState function).

database/database_test.go
```diff
@@ -177,6 +178,41 @@ func TestStateForDomain(t *testing.T) {
...
+
+	domainStates, err := testDB.StatesForDomains([]string{"a.test", "b.test"})
+	if err != nil {
+		t.Errorf("Cannot get states for test StatesForDomains: %s", err)
+		return
+	}
+
+	sort.Slice(domainStates, func(i, j int) bool { return domainStates[i].Name < domainStates[j].Name })
+	if len(domainStates) != 2 {
+		count := strconv.Itoa(len(domainStates))
```
carlosjoan91, line 201: Since this is only used for the printout, you can instead do t.Errorf("Inccorect count of states for test StatesForDomains %d", len(domainStates))

Or better still, save the len(domainStates) in a "got" variable, then check against that (and the literal "2" in a "want" variable), then you can change the error message so it includes something like t.Errorf("Incorrect count of states for test StatesForDomains, got %d, wanted %d", got, want)

database/database_test.go
```diff
@@ -177,6 +178,41 @@ func TestStateForDomain(t *testing.T) {
...
+		t.Errorf("Inccorect count of states for test StatesForDomains %s", count)
+	}
+	if domainStates[0].Name != "a.test" || domainStates[1].Name != "b.test" {
+		t.Errorf("Domain states not populated correctly for test StatesForDomains")
```
carlosjoan91, line 205: Consider adding the got/want information in this error too so a failing test is easier to debug.

database/database_test.go
```diff
@@ -177,6 +178,41 @@ func TestStateForDomain(t *testing.T) {
...
+	}
+
+	// tests domain not in database
+	domainStates, err = testDB.StatesForDomains([]string{"a.test", "b.test", "c.test"})
+	wantError := "datastore: no such entity"
+	if err.Error() != wantError {
+		t.Errorf("Non-preloaded domain treated incorrectly for test StatesForDomains %s", err)
```
carlosjoan91, line 212: Also consider adding got/want error information here

database/database_test.go
```diff
@@ -239,6 +275,49 @@ func TestStatesWithStatus(t *testing.T) {
...
+		{Name: "pending.test", Status: StatusPending},
+		{Name: "pendingremoval.test", Status: StatusPendingRemoval},
+		{Name: "rejected.test", Status: StatusRejected},
+		{Name: "removed.test", Status: StatusRemoved},
+		{Name: "unknown.test", Status: StatusUnknown},
+		{Name: "pendingautomatedremoval.test", Status: StatusPendingAutomatedRemoval}}
+
+	for _, state := range testStates {
+		if err := testDB.PutState(state); err != nil {
+			t.Errorf("cannot put state %s", err)
```
carlosjoan91, line 292: This should also be a t.Fatalf (since you are also not intending to test PutState here).

database/database_test.go
```diff
@@ -239,6 +275,49 @@ func TestStatesWithStatus(t *testing.T) {
...
+	testDomains := []string{"preloaded.test", "pending.test", "pendingremoval.test", "rejected.test", "removed.test", "unknown.test", "pendingautomatedremoval.test"}
+	var statuses []string
+	statusReport := func(format string, args ...interface{}) {
+		formatted := fmt.Sprintf(format, args...)
+		statuses = append(statuses, formatted)
+	}
+
+	err := testDB.SetPendingAutomatedRemoval(testDomains, statusReport)
+	if err != nil {
+		t.Errorf("Error for test SetAutoamtedPendingRemoval: %s", err)
```
carlosjoan91, line 306: nit: typo in SetAutomatedPendingRemoval (also consider making this a bit more descriptive (e.g. "Unexpected error in SetAutomatedPendingRemoval").

carlosjoan91 (commented):
Thanks! Left a question about a block of code, and a bunch of smaller/nit comments.

database/database.go
tKabiawu: I left out a part in this section. If len(keys) exceeded batch size, then I was going to store the output of the call to getDomainStates in a slice. Then, when we call getDomainStates on the remaining keys at the end of the function, I was going append this output to the slice from earlier. Do you think that this is neccessary?

tKabiawu: Done

tKabiawu: Just updated the PR with the change I proposed

database/database_test.go
tKabiawu: Done.

database/database_test.go
tKabiawu: Done

database/database_test.go
tKabiawu: Done

database/database_test.go
tKabiawu: Done

database/database_test.go
tKabiawu: Done.

database/database_test.go
tKabiawu: Done

database/database_test.go
```diff
@@ -30,7 +29,7 @@ func ExampleTempLocalDatabase() {
 func TestMain(m *testing.M) {
 	localDatabase, shutdown, err := TempLocalDatabase()
 	if err != nil {
-		fmt.Fprintf(os.Stderr, "could not initialize local backend %s", err)
+		fmt.Fprintf(os.Stderr, "could not initialize local backend because %s", err)
```
carlosjoan91, line 32: Very minor styling nit: I think the most common way to print this out is just add a : between your explanation and the error (so "could not initialize local backend: %s" in this case)

database/database_test.go
```diff
@@ -182,11 +181,11 @@ func TestStatesForDomains(t *testing.T) {
 	resetDB()
 
 	if err := testDB.PutState(DomainState{Name: "a.test", Status: StatusPreloaded}); err != nil {
-		t.Errorf("cannot put state %s for test StatatesForDomains", err)
+		t.Fatalf("Cannot put state a.test for test StatatesForDomains")
```
carlosjoan91, line 184: Here and in the next one: Printint the error is fine in the Fatalf too (maybe at the end "cannot put state a.test for test StatesForDomains: %s", err)

database/database_test.go
```diff
@@ -197,19 +196,21 @@ func TestStatesForDomains(t *testing.T) {
...
 	}
+
 	if domainStates[0].Name != "a.test" || domainStates[1].Name != "b.test" {
-		t.Errorf("Domain states not populated correctly for test StatesForDomains")
+		t.Errorf("Domain states not populated correctly for test StatesForDomains. Got the names %s and %s, but wanted a.test and b.test", domainStates[0].Name, domainStates[1].Name)
 	}
 
 	// tests domain not in database
-	domainStates, err = testDB.StatesForDomains([]string{"a.test", "b.test", "c.test"})
+	domainStates, getErr := testDB.StatesForDomains([]string{"a.test", "b.test", "c.test"})
```
carlosjoan91, line 210: nit: gotErr

database/database.go
```diff
@@ -161,13 +162,14 @@ func (db DatastoreBacked) StatesForDomains(domains []string) (states []DomainSta
...
 	if datastoreErr != nil {
-		return states, datastoreErr
+		return nil, datastoreErr
 	}
 
 	getDomainStates := func(keys []*datastore.Key) ([]DomainState, error) {
 		domainStates := make([]DomainState, len(keys))
-		if err := client.GetMulti(c, keys, domainStates); err != nil {
-			return nil, err
+		getErr := client.GetMulti(c, keys, domainStates)
```
carlosjoan91, line 170: The "gotX" naming convention is mostly for tests, I think err is probably better here

carlosjoan91: (and also what you had before of setting the err inside the if statement)

database/database.go
```diff
@@ -153,6 +154,49 @@ func (db DatastoreBacked) StateForDomain(domain string) (state DomainState, err
...
+		}
+		for i := range domainStates {
+			domainStates[i].Name = keys[i].Name
+		}
+		return domainStates, nil
+	}
+
+	var keys []*datastore.Key
+	var res []DomainState
+	var getErr error
```
carlosjoan91, line 182: same comment about using err here

database/database.go
```diff
@@ -153,6 +154,49 @@ func (db DatastoreBacked) StateForDomain(domain string) (state DomainState, err
...
+	for _, domain := range domains {
+		key := datastore.NameKey(domainStateKind, domain, nil)
+		keys = append(keys, key)
+		if len(keys) >= batchSize {
+			tempStates, getErr := getDomainStates(keys)
+			res = append(res, tempStates...)
+			if getErr != nil {
```
carlosjoan91, line 189: Nit: Prefer doing the error check immediately after the call to getDomainStates (since the append call doesn't matter if we get an error)

database/database.go
carlosjoan91: (and I think you can omit this declaration since the error is always used only in the block it is received).

database/database.go
```diff
@@ -272,3 +316,52 @@ func (db DatastoreBacked) DeleteIneligibleDomainStates(domains []string) (err er
...
+
+	return states, nil
+}
+
+// SetPendingAutomatedRemoval sets the status of a list of domains to StatusPendingAutoamtedRemoval
+func SetPendingAutomatedRemoval(db Database, domains []string, logf func(fomat string, args ...interface{})) error {
+	setDomainStates := func(domainStates []DomainState) error {
+		logf("Updating %d entries...", len(domainStates))
+
+		if err := db.PutStates(domainStates, blackholeLogf); err != nil {
```
carlosjoan91, line 350: Since your function receives a logf parameter, I think you want to pass that in here instead of unconditionally using blackholeLogF

database/database.go
```diff
@@ -272,3 +316,52 @@ func (db DatastoreBacked) DeleteIneligibleDomainStates(domains []string) (err er
...
+			logf(" faild.\n")
```
carlosjoan91, line 351: nit: Typo for failed?

database/database.go
```diff
@@ -359,3 +316,52 @@ func (db DatastoreBacked) DeleteIneligibleDomainStates(domains []string) (err er
 	}
 	return delete(keys)
 }
+
+// GetAllIneligibleDomainStates returns all the ineligible domains in the database
+func (db DatastoreBacked) GetAllIneligibleDomainStates() (states []IneligibleDomainState, err error) {
```
carlosjoan91, line 321: Do you think this function can be split into a separate PR? I don't think it interacts with the rest of this PR

carlosjoan91 (commented):
Thanks! For the most part this looks good, left a few (mostly styling related) comments.

database/database_test.go
```diff
@@ -301,9 +302,9 @@ func TestSetAutomatedPendingRemoval(t *testing.T) {
 		statuses = append(statuses, formatted)
 	}
 
-	err := testDB.SetPendingAutomatedRemoval(testDomains, statusReport)
+	err := SetPendingAutomatedRemoval(testDB, testDomains, statusReport)
 	if err != nil {
-		t.Errorf("Error for test SetAutoamtedPendingRemoval: %s", err)
+		t.Errorf("Unexpected error for test SetAutoamtedPendingRemoval: %s", err)
```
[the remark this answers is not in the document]
tKabiawu: Done

database/database_test.go
tKabiawu: Done

database/database_test.go
tKabiawu: Done

database/database.go
tKabiawu: Done

database/database.go
tKabiawu: Done

database/database.go
tKabiawu: Done

database/database.go
tKabiawu: I'm not sure what's wrong with Github's tracking. This function was already in the database.go file -- the only functions I added were StatesForDomains and SetPendingAutomatedRemoval(). Not really sure how to fix this, so would it be a problem if it was included in this pr?

database/database.go
```diff
@@ -153,6 +154,50 @@ func (db DatastoreBacked) StateForDomain(domain string) (state DomainState, err
...
+
+	var keys []*datastore.Key
+	var res []DomainState
+	for _, domain := range domains {
+		key := datastore.NameKey(domainStateKind, domain, nil)
+		keys = append(keys, key)
+		if len(keys) >= batchSize {
+			if tempStates, err := getDomainStates(keys); err != nil {
+				return nil, err
+			} else {
```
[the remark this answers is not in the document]
carlosjoan91: This is because tempstates is declared and assigned in the if statement, you can declare tempStates one line earlier if you want to remove the else statement, e.g.:
```
if len(keys) >= batchSize {
    var tempStates []DomainState
    if tempStates, err := getDomainStates(keys); err != nil {
```

database/database.go
```diff
@@ -167,9 +167,8 @@ func (db DatastoreBacked) StatesForDomains(domains []string) (states []DomainSta
 
 	getDomainStates := func(keys []*datastore.Key) ([]DomainState, error) {
 		domainStates := make([]DomainState, len(keys))
-		getErr := client.GetMulti(c, keys, domainStates)
-		if getErr != nil {
-			return nil, getErr
+		if dsErr := client.GetMulti(c, keys, domainStates); dsErr != nil {
```
carlosjoan91, line 170: minor styling nit: this can just be called err, since it's only used here and there is no risk of confusing with another err.

carlosjoan91 (approved):
Thanks! This LGTM with a small nit, and an optional comment (responding to your comment).

database/database.go
tKabiawu: Done

PR #239, opened 2023-08-08
Configured cron.yaml to hit RemoveInelligibleDomains endpoint

tKabiawu:


api/domain_handlers_test.go
```diff
@@ -98,6 +98,8 @@ func TestAddIneligibleDomain(t *testing.T) {
 	if err != nil {
 		t.Fatalf("[%s] %s", "NewRequest Failed", err)
 	}
+	r.Header.Set("X-Appengine-Cron", "true")
+	r.RemoteAddr = "0.1.0.2"
```
nharper, line 102: nit: make a helper function that returns a properly configured request for this endpoint (instead of repeating this for each occurrence)

api/domain_handlers.go
```diff
@@ -387,8 +387,12 @@ func (api API) Remove(w http.ResponseWriter, r *http.Request) {
 // database and change the status to PendingAutomatedRemoval if the domain
 // does not follow the requirements for more than 2 crawls
 
-// Example: GET /removeineligibledomains?
+// Example: GET /remove-ineligible-domains
 func (api API) RemoveIneligibleDomains(w http.ResponseWriter, r *http.Request) {
+	// ensures endpoint requests can only come from App Engine
+	if r.Header.Get("X-Appengine-Cron") != "true" || r.RemoteAddr != "0.1.0.2" {
+		return
```
nharper, line 394: Before returning, at least set a status code on the response indicating that the request won't be processed (e.g. a 403).

Also, add a test case that if the header or ip address is missing, the status code gets returned.

cron.yaml
```diff
@@ -0,0 +1,6 @@
+cron: 
+- description: "Removes inneligible domains"
+  url: "/api/v2/remove-ineligible-domains"
+  schedule: every monday 09:00
+  timezone: america/new_york
```
nharper, line 5: To answer the question about deploying this: can you add a change to the Makefile to run `gcloud app deploy cron.yaml` as part of the deploy step (similar to what's done for app.yaml)?

api/domain_handlers.go
tKabiawu: Done

cron.yaml
tKabiawu: Done

Makefile
```diff
@@ -17,6 +17,7 @@ pre-commit: lint build test
 deploy: version
 	date
 	time gcloud app deploy app.yaml
+	gcloud app deploy cron.yaml
```
tKabiawu, line 20: not sure if I need the time here

nharper: I haven't found the output from `time` useful, so I'm fine skipping that.

api/domain_handlers.go
nharper: nit: use a constant (`http.StatusForbidden`) instead of a hardcoded value

api/domain_handlers_test.go
```diff
@@ -98,12 +104,19 @@ func TestAddIneligibleDomain(t *testing.T) {
 	if err != nil {
 		t.Fatalf("[%s] %s", "NewRequest Failed", err)
 	}
+	// tests that a non-App Engine http request returns a 403 status code
+	invalidW := httptest.NewRecorder()
+	invalidW.Body = &bytes.Buffer{}
+	if api.RemoveIneligibleDomains(invalidW, r); invalidW.Code != 403 {
+		t.Errorf("HTTP Response INvalid: Status code for invalid http request is not 403")
```
nharper, line 111: nit: "invalid"

tKabiawu: Done

api/domain_handlers.go
tKabiawu: Done

api/domain_handlers_test.go
nharper: the variable name was fine - the nit was about the string that has "INvalid"

tKabiawu: Ah okay, now Done

nharper (approved):

PR #240, opened 2023-08-08
Update golang runtime

nharper:
Update app.yaml to use the latest golang runtime for app engine flex, which is currently go 1.20. (This change also removes a no longer needed environment variable.)


carlosjoan91 (approved):
LGTM

PR #248, opened 2023-10-31
Handle pending-automated-removal in UI

nharper:


nharper:
This is the first introduction of [Template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) in the frontend code. From looking at https://caniuse.com/template-literals, I'm not concerned about breakage.

jdeblasio (approved):

PR #250, opened 2023-11-03
Support transition from pending-automated-removal to pending

nharper:
If a domain has the pending-automated-removal status (from failing to meet the requirements) but is then fixed so that it meets the requirements before the automated removal gets processed, it should be able to transition out of the pending-automated-removal state.

### DIFF
--- a/api/api_test.go
+++ b/api/api_test.go
@@ -210,13 +210,6 @@ func TestAPI(t *testing.T) {
 				Warnings: []hstspreload.Issue{{Code: "server.preload.already_pending"}},
 			}}},
 
-		// pending automated removal
-		{"pending automated removal", data1, failNone, api.PendingAutomatedRemoval, "GET", "",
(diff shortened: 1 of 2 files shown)


carlosjoan91 (approved):

nharper:
I don't think this fix makes sense (and I think the existing behavior for StatusPendingRemoval also doesn't make sense).

If a domain is in StatusPendingAutomatedRemoval (or StatusPendingRemoval), that means that the domain is currently on the preload list, but it had met the criteria for removal and was about to be removed from the list. If the domain passes the checks of the submit endpoint, that means that it no longer meets the removal criteria. Since the domain is already on the list, it shouldn't now be pending submission to the list, it should just be StatusPreloaded.

I could also see the case for merging this PR so that StatusPendingAutomatedRemoval behaves the same as StatusPendingRemoval, and leave it to a future issue for whether domains pending automated removal should transition to a different state from the submit endpoint.

nharper:
There's more going on here. Filed issue #251 to pick up this work.

PR #260, opened 2024-09-11
Fix Update endpoint

nharper:
Multiple state transitions in the Update endpoint were broken. This fixes those transitions and deprecates the Rejected status in favor of Removed.

### DIFF
--- a/api/update.go
+++ b/api/update.go
@@ -8,21 +8,6 @@ import (
 	"github.com/chromium/hstspreload/chromium/preloadlist"
 )
 
-func difference(from []preloadlist.Entry, take []preloadlist.Entry) (diff []preloadlist.Entry) {
-	takeSet := make(map[string]bool)
-	for _, elem := range take {
-		takeSet[elem.Name] = true
-	}
-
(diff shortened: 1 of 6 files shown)


api/update.go
```diff
@@ -40,73 +25,73 @@ func (api API) Update(w http.ResponseWriter, r *http.Request) {
...
+	updated := 0
+	removed := 0
+	for _, entry := range preloadList.Entries {
+		if entry.Mode != preloadlist.ForceHTTPS {
+			continue
+		}
+		domainState, found := domainStates[entry.Name]
+		if !found {
+			// entry is on the preload list but not marked as
+			// preloaded or pending removal in the database. Mark it
```
carlosjoan91, line 61: Nit: preloaded, pending removal, or pending automated removal (for completeness sake)

carlosjoan91 (approved):
Approved with a very small nit about a comment. Thanks for the fix

api/update.go
nharper: Done.

PR #261, opened 2024-09-11
Fix lint errors on main

nharper:


nharper:
I missed this when merging #237.

PR #264, opened 2025-01-22
Reword content on hstspreload.org

nharper:
Many web browsers will upgrade HTTP navigations to HTTPS by default now, which lessens the value proposition of preloading HSTS (vs HSTS without preloading). Because of this, we don't recommend that sites preload their domains.

Part of this change also modifies the styling of the submission form to have elements generally sized closer to body text size and keep text left-justified instead of centered, to match the rest of the page.

### DIFF
--- a/frontend/index.html
+++ b/frontend/index.html
@@ -23,81 +23,87 @@
 <body class="theme-green">
 <a class="github-fork-ribbon" href="https://github.com/chromium/hstspreload.org" data-ribbon="On GitHub" title="On GitHub">On GitHub</a>
 
-<div class="content form">
-  <form id="domain-form" class="hidden">
-    <h2><label for="domain">
-      Enter a domain:
-    </label></h2>
+<div class="content">
+  <section id="hsts">
+    <h2><a class="hash-link" href="#hsts">HTTP Strict Transport Security (HSTS)</a></h2>
+    <p>
+      <a href="https://en.wikipedia.org/wiki/HTTP_Strict_Transport_Security">HTTP Strict Transport Security (HSTS)</a> is a mechanism for websites to instruct web browsers that the site should only be accessed over HTTPS. This mechanism works by sites sending a <code>Strict-Transport-Security</code> HTTP response header containing the site's policy.
+    </p>
 
-    <input id="domain" name="domain" type="text" placeholder="example.com"
-      autocorrect="off" autocapitalize="off" spellcheck="false">
-    <br>
-    <input id="check" type="submit" value="Check HSTS preload status and eligibility">
+    <p>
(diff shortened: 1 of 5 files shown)


frontend/index.html
```diff
@@ -151,65 +157,78 @@ <h2><a class="hash-link" href="#continued-requirements">Continued Requirements</
...
+  <section id="submission-form" class="form">
+    <h2><a class="hash-link" href="#submission-form">Submission Form</a></h2>
     <p>
-      Consult the <a href="https://wiki.mozilla.org/Security/Guidelines/Web_Security">Mozilla Web Security guidelines</a> and the <a href="https://developers.google.com/web/fundamentals/security/?hl=en">Google Web Fundamentals pages on security</a> for more concrete advice about HTTPS deployment.
+      If you still wish to submit your domain for inclusion in Chrome's HSTS preload list and you have followed our <a href="#deployment-recommendations">deployment recommendations</a> of slowly ramping up the <code>max-age</code> of your site's <code>Strict-Transport-Security</code> header, you can use this form to do so:
     </p>
 
+    <form id="domain-form" action="#submission-form" class="hidden">
+      <p>
+        <label for="domain">Domain to preload:</label>
```
lgarron, line 180: Teensy nit, but this throws off the centering, something like this would easily keep it centered:
```suggestion
        <hr>
        <label for="domain">Domain to preload:</label>
        <br>
```

nharper:
Here's what the form looks like without the added `<hr>` and `<br>` tags:
Screenshot 2025-01-23 at 12 41 59 PM

With the added `<hr>` and `<br>` tags:
Screenshot 2025-01-23 at 12 41 48 PM

I slightly prefer the former, but @jdeblasio if you prefer the latter I'm happy to change it. (Note that in neither case is the form centered - that's one of the CSS changes made here.)

jdeblasio:
Ew. Yeah I don't like the <hr> after all. The first option is fine.

lgarron:
Ah, I think I was looking at it on a screen and thus seeing it wrapped:

clipboard

I think it's still relatively clear how to use, though. And I certainly appreciate the other content changes!

deviant:
Hi, I realize this change was made a while ago, but could somebody clarify whether dropping the recommendation to set `max-age` to 2 years was intentional? I remembered this being the case from several years ago, and also found your recommendation referenced from other websites when trying to track this down again.

The same header example is still listed on the current version of the site, but that now only has "an example of a valid HSTS header" to say about it, rather than directly suggesting people use it when seeking to preload their sites. I liked there being a recommended header value to use (rather than one being put forward as an arbitrary sample), since it made deciding what to pick very simple and straightforward.

deviant:
Thanks both, these are insightful responses (and the Chromium thread made for an interesting read as well!).

PR #275, opened 2025-05-16
Update dependencies

jub0bs:
go.mod and go.sum were updated by running `go get -u && go mod tidy`. Since the update to go.mod included an update to the go version to 1.23.0, this change also updates the CI config to the same version.


jub0bs:
Note: I've updated to Go 1.23 because it's currently [the oldest version of Go that's still supported](https://go.dev/doc/devel/release#policy).

PR #279, opened 2025-08-08
Process /api/v2/remove-ineligible-domains in shards

nharper:
Processing all domains at once is failing to complete. Adding query parameters to the endpoint to have it process a subset of the domains on each run should result in each run using less resources and taking less time. The cron configuration is changed to split the work into 4 approximately evenly sized shards.


jdeblasio (commented):
Thanks!

jdeblasio (approved):
Belated thanks!

